PR #1, opened 2016-11-23
Middlewares

tomusdrw:
Generic `IoHandler` allowing to write transport-level middlewares.

Example:
HTTP server can extract metadata from the request and then augment RPC call (for instance alter passed parameters)

Please review @debris


http/src/handler.rs
```diff
@@ -18,12 +18,38 @@ pub struct PanicHandler {
...
+
+impl RpcHandler for IoHandler {
+	type Metadata = ();
+
+	fn handle_request<H>(&self, request_str: &str, response_handler: H, _meta: Option<Self::Metadata>) where
+		H: ResponseHandler<Option<String>, Option<String>> + 'static
+	{
+		GenericIoHandler::handle_request(self, request_str, response_handler)
+	}
+}
```
debris: this file has incorrect formatting

core/src/io.rs
```diff
@@ -229,75 +268,68 @@ impl IoHandler {
...
+		&self.request_handler
 	}
 
-	fn handle<H: ResponseHandler<Option<String>, Option<String>> + 'static>(&self, request_str: &str, response_handler: H, session: Option<Session>) {
-		trace!(target: "rpc", "Request: {} in session.", request_str);
-
-		let handler = Handler::new(
+	/// Converts the handler to serializing one.
+	pub fn convert_handler<H>(response_handler: H) -> Handler<Option<String>, Option<Response>, Notification> where
+		H: ResponseHandler<Option<String>, Option<String>> + 'static
```
tomusdrw: Did it for `Handler<Option<String> ....>`, but can't do it for `ResponseHandler<Option<String>, Option<String>>`:
```
note: type aliases cannot be used for traits
```

http/src/handler.rs
tomusdrw: The formatting is consistent with the whole `http` crate:
https://github.com/ethcore/jsonrpc/blob/master/http/.editorconfig

Formatted using tabs, but probably Github is taking into account `tab_width` defined in `.editorconfig`, I can make a separate PR to unify formatting across all crates.

debris: 👍

PR #13, opened 2016-12-10
Share `Trailing` struct

svyatonik:


debris:
@rphmeier could we replace `Trailing<T>` with `Option<T>`? It would require manual deserialization, but imo it's more flexible and clear

Currently it's impossible to create a function with`Trailing<bool>` set to `true` by default

tomusdrw:
`Option<T>` has different semantics: `["a", "b", null]` vs `Trailing<T>: ["a", "b"]` I think it would be better to keep it this way.
That said, I agree that making `Trailing` more like `Option` were the `Default` is not required would be a bit more convenient.

rphmeier:
No, I don't think `Option<T>` will compile because it has a `Deserialize` implementation; the whole point is to introduce a structure which is known not to be deserializable so the compiler can infer that `impl Foo for X<A, B, C, Trailing<D>>` doesn't conflict with `impl Foo for X<A, B, C, D>`.

rphmeier:
That said, the PR LGTM. Merge at will :)

PR #22, opened 2016-12-14
Macros for methods with Metadata

### LINKED ISSUE #20: Metadata support for Macros

tomusdrw:
Closes #20


core/src/io.rs
```diff
@@ -157,32 +149,40 @@ impl<T: Metadata> MetaIoHandler<T> {
 
 /// Simplified `IoHandler` with no `Metadata` associated with each request.
 #[derive(Default)]
-pub struct IoHandler(MetaIoHandler<()>);
+pub struct IoHandler<M: Metadata = ()>(MetaIoHandler<M>);
```
debris: why did you change that?

macros/src/auto_args.rs
```diff
@@ -101,6 +147,11 @@ pub trait WrapAsync<B: Send + Sync + 'static> {
 	fn wrap_rpc(&self, base: &B, params: Params) -> BoxFuture<Value, Error>;
 }
 
+/// Wrapper trait for meta RPC functions.
+pub trait WrapMeta<B: Send + Sync + 'static, M: Metadata> {
+	fn wrap_rpc(&self, base: &B, params: Params, meta: M) -> BoxFuture<Value, Error>;
```
rphmeier: limitation here is that all metadata-taking methods must be async?

core/src/io.rs
tomusdrw: To support writing test cases not using metadata easier:

```rust
assert_eq!(io.handle_request_sync(&request), Some(response));
```

and for meta you need to do:
```rust
assert_eq!((*io).handle_request_sync(&request, meta), Some(response));
```
or just use `MetaIoHandler` directly.

macros/src/auto_args.rs
tomusdrw: Yes. Do you think we should support sync meta too?

rphmeier: IMO it should all be async now that futures are integrated. But I suppose this limitation is fine.

PR #25, opened 2016-12-19
initial jsonrpc v1 support

debris:
@tomusdrw if you agree this design I will add some tests

### DIFF
--- a/core/src/io.rs
+++ b/core/src/io.rs
@@ -175,7 +175,7 @@ impl<T: Metadata> MetaIoHandler<T> {
 				let valid_version = self.compatibility.is_version_valid(jsonrpc);
 
 				let result = match (valid_version, self.methods.get(&method.method)) {
-					(false, _) => futures::failed(Error::invalid_request()).boxed(),
+					(false, _) => futures::failed(Error::invalid_version()).boxed(),
 					(true, Some(&RemoteProcedure::Method(ref method))) => method.call(params, meta),
 					(true, _) => futures::failed(Error::method_not_found()).boxed(),
 				};
--- a/core/src/io.rs
+++ b/core/src/io.rs
@@ -5,7 +5,7 @@ use serde_json;
 use futures::{self, Future, BoxFuture};
 
 use calls::{RemoteProcedure, Metadata, RpcMethodSync, RpcMethodSimple, RpcMethod, RpcNotificationSimple, RpcNotification};
-use types::{Params, Error, ErrorCode};
+use types::{Params, Error, ErrorCode, Version};
 use types::{Request, Response, Call, Output};
 
 fn read_request(request_str: &str) -> Result<Request, Error> {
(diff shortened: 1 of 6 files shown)


core/src/io.rs
```diff
@@ -17,13 +17,52 @@ fn write_response(response: Response) -> String {
...
+/// By default compatible only with jsonrpc v2
 #[derive(Default)]
 pub struct MetaIoHandler<T: Metadata> {
+	compatibility: Compatibility,
 	methods: HashMap<String, RemoteProcedure<T>>,
 }
 
 impl<T: Metadata> MetaIoHandler<T> {
+	/// Creates new `MetaIoHandler` without any metadata compatible with specified protocol version.
+	pub fn new(compatibility: Compatibility) -> Self {
```
tomusdrw: I would leave `new()` as is and add new `with_compatibility()` method.

core/src/io.rs
```diff
@@ -17,13 +17,52 @@ fn write_response(response: Response) -> String {
...
+///
+/// By default compatible only with jsonrpc v2
 #[derive(Default)]
 pub struct MetaIoHandler<T: Metadata> {
+	compatibility: Compatibility,
 	methods: HashMap<String, RemoteProcedure<T>>,
 }
 
 impl<T: Metadata> MetaIoHandler<T> {
+	/// Creates new `MetaIoHandler` without any metadata compatible with specified protocol version.
```
tomusdrw: Comment is invalid (it has metadata)

core/src/io.rs
```diff
@@ -121,10 +160,12 @@ impl<T: Metadata> MetaIoHandler<T> {
 				let params = method.params.unwrap_or(Params::None);
 				let id = method.id;
 				let jsonrpc = method.jsonrpc;
+				let valid_version = self.compatibility.is_version_valid(jsonrpc);
 
-				let result = match self.methods.get(&method.method) {
-					Some(&RemoteProcedure::Method(ref method)) => method.call(params, meta),
-					_ => futures::failed(Error::new(ErrorCode::MethodNotFound)).boxed(),
+				let result = match (valid_version, self.methods.get(&method.method)) {
+					(false, _) => futures::failed(Error::invalid_request()).boxed(),
```
tomusdrw: Would add more descriptive error for this.

core/src/io.rs
```diff
@@ -157,6 +202,11 @@ impl IoHandler {
 	pub fn new() -> Self {
 		IoHandler::default()
 	}
+
+	/// Creates new `IoHandler` without any metadata compatible with specified protocol version.
+	pub fn new_compatible_with(compatibility: Compatibility) -> Self {
```
tomusdrw: IMHO `with_compatbility` would be enough

core/src/types/request.rs
```diff
@@ -11,7 +11,7 @@ use super::{Id, Params, Version, Value};
 pub struct MethodCall {
 	/// A String specifying the version of the JSON-RPC protocol.
 	/// MUST be exactly "2.0".
```
tomusdrw: Comment should be updated.

core/src/types/request.rs
```diff
@@ -29,7 +29,7 @@ pub struct MethodCall {
 pub struct Notification {
 	/// A String specifying the version of the JSON-RPC protocol.
 	/// MUST be exactly "2.0".
```
tomusdrw: Comment should be updated.

core/src/types/request.rs
```diff
@@ -233,20 +233,22 @@ fn request_deserialize_batch() {
...
 		Call::Notification(Notification {
-			jsonrpc: Version::V2,
+			jsonrpc: Some(Version::V2),
 			method: "update".to_owned(),
 			params: Some(Params::Array(vec![Value::U64(1)]))
 		})
 	]))
 }
 
+// TODO: fix this test. it's valid now, since jsonrpc v1 does not require "version" field
```
tomusdrw: Indeed. (Perhaps add some unexpected field instead)

core/src/types/response.rs
```diff
@@ -56,7 +56,7 @@ impl Output {
 	pub fn invalid_request(id: Id) -> Self {
```
tomusdrw: This should probably get `Option<Version>` as well.

core/src/types/response.rs
```diff
@@ -115,7 +115,7 @@ impl From<Error> for Response {
 	fn from(error: Error) -> Self {
```
tomusdrw: This implementation is no longer valid. Should be more explicity: `Response::from_error(error, jsonrpc)`

core/src/types/version.rs
```diff
@@ -3,7 +3,7 @@ use serde::{Serialize, Serializer, Deserialize, Deserializer, Error};
 use serde::de::Visitor;
 
 /// Protocol Version
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Debug, PartialEq, Clone, Copy)]
 pub enum Version {
```
tomusdrw: Shouldn't this contain other versions? (Does spec say if `{"jsonrpc":"1.0", "method"....}` is a valid request?)

core/src/io.rs
debris: there is no `new` currently, so I will add it ;)

PR #29, opened 2017-01-17
TCP migration to tokio, part I

NikVolf:


tcp/src/tests.rs
```diff
@@ -14,96 +14,91 @@
...
+    use tokio_core::io;
+    use futures::{Future, future};
+
+    ::logger::init_log();
+    let addr: SocketAddr = "127.0.0.1:17780".parse().unwrap();
+    let server = casual_server(&addr);
+    ::std::thread::spawn(move || server.run().expect("Server must run with no issues"));
+    wait(100);
+
+    let mut core = Core::new().expect("Tokio Core should be created with no errors");
```
NikVolf: Pretty much shows that request handling is working

tcp/src/line_codec.rs
```diff
@@ -0,0 +1,69 @@
...
+    type In = String;
+    type Out = String;
+
+    fn decode(&mut self, buf: &mut EasyBuf) -> io::Result<Option<Self::In>> {
+        if let Some(i) = buf.as_slice().iter().position(|&b| b == b'\n') {
+            let line = buf.drain_to(i);
+            buf.drain_to(1);
+            return match str::from_utf8(&line.as_ref()) {
+                Ok(s) => Ok(Some(s.to_string())),
+                Err(_) => Err(io::Error::new(io::ErrorKind::Other,
```
tomusdrw: A bit weird line break, would keep it in a single line.

tcp/src/line_codec.rs
```diff
@@ -0,0 +1,69 @@
...
+use tokio_core::io::{Codec, EasyBuf};
+
+pub struct LineCodec;
+
+impl Codec for LineCodec {
+    type In = String;
+    type Out = String;
+
+    fn decode(&mut self, buf: &mut EasyBuf) -> io::Result<Option<Self::In>> {
+        if let Some(i) = buf.as_slice().iter().position(|&b| b == b'\n') {
```
tomusdrw: I think it may break if JSON contains a string with a line break:
`{"test": "something\\nasd"}`

tcp/src/meta.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use std::net::SocketAddr;
+use jsonrpc::Metadata;
+
+pub struct RequestContext {
+    pub peer_addr: SocketAddr,
+}
+
+pub trait MetaExtractor<M: Metadata> : Send + Sync {
+    fn extract(&self, context: &RequestContext) -> Option<M>;
```
tomusdrw: `Metadata` is `Default` so I think there is no need here to return `Option<M>` (becaue we will do `unwrap_or_default()` anyway), let's just force the user to return something instead.

tcp/src/meta.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+pub struct NoopExtractor;
+
+impl<M: Metadata> MetaExtractor<M> for NoopExtractor {
+    fn extract(&self, _context: &RequestContext) -> Option<M> { None }
+}
+
+#[derive(Clone)]
+pub struct SocketMetadata {
```
tomusdrw: I think `SocketMetadata` and `PeerMetaExtractor` shouldn't be implement here, I would implement those in the app using that library.

tcp/src/server.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn new(addr: SocketAddr, handler: Arc<MetaIoHandler<M>>) -> Self {
+        Server { listen_addr: addr, handler: handler, meta_extractor: Arc::new(NoopExtractor)}
+    }
+
+    pub fn extractor(mut self, meta_extractor: Arc<MetaExtractor<M>>) -> Self {
+        self.meta_extractor = meta_extractor;
+        self
+    }
+
+    pub fn run(&self) -> std::io::Result<()> {
```
tomusdrw: Let's consume the Server, here it cannot be used twice anyway

tcp/src/tests.rs
```diff
@@ -14,96 +14,127 @@
...
+#[test]
+fn doc_test() {
+    ::logger::init_log();
+
+    let mut io = MetaIoHandler::<()>::new();
+    io.add_method("say_hello", |_params| {
+        Ok(Value::String("hello".to_string()))
+    });
+    let server = Server::new(SocketAddr::from_str("0.0.0.0:17770").unwrap(), Arc::new(io));
+    ::std::thread::spawn(move || server.run().expect("Server must run with no issues"));
```
tomusdrw: just `use std::thread` somewhere above, cause it's used couple of times.

tcp/src/lib.rs
```diff
@@ -28,486 +28,41 @@
...
+//! 	let server = Server::new(SocketAddr::from_str("0.0.0.0:9993").unwrap(), Arc::new(io));
+//!     ::std::thread::spawn(move || server.run().expect("Server must run with no issues"));
 //! }
 //! ```
 
-extern crate jsonrpc_core;
+extern crate jsonrpc_core as jsonrpc;
 #[macro_use] extern crate log;
 #[macro_use] extern crate lazy_static;
 extern crate env_logger;
```
tomusdrw: Should be used only in tests (maybe even do `extern crate env_logger` in `mod logger`)

tcp/src/lib.rs
```diff
@@ -28,486 +28,41 @@
...
+#[cfg(test)] mod tests;
+
+pub use server::Server;
+pub use meta::{MetaExtractor, PeerMetaExtractor, SocketMetadata};
+
+/// Default JSON RPC over TCP/IP server configuration
+pub fn server(listen_addr: &::std::net::SocketAddr, io: jsonrpc::MetaIoHandler<SocketMetadata>) -> Server<SocketMetadata> {
+    Server::new(
+        listen_addr.clone(),
+        ::std::sync::Arc::new(io))
```
tomusdrw: Let's just import that `Arc`, it's used in 3 places

tcp/src/line_codec.rs
NikVolf: It should indeed break there, and it's not a bug
it's client that must be aware of such behaviour and escape requests

in LineCodec, `\n` supersedes any json formatting, it's the only way to split requests for it

tomusdrw: But the new line is already escaped, It's a valid JSON request that could be sent over HTTP without a trouble. How could I escape that further?

You can try to parse incoming requests for correctness, or introduce another delimiter that cannot be part of any JSON (like `\n\n`)

tomusdrw: Nevermind, just noticed that it's looking for `\n` character not string so it's fine!

tcp/src/meta.rs
NikVolf: why so?
I thought of the default `SocketMetadata` which the user can replace with custom metadata
Won't the converntion over configuration be better here?

tomusdrw: Default metadata is `()`, most of users won't need any `Metadata` and they can use simplified types like `IoHandler`.

Right now there is not much data you can extract from `RequestContext`, but just as a general rule I don't think it suits here. In principle `Metadata` can contain transport-specific data but the struct itself should be transport-agnostic, so that I can extract same `Metadata` object from different transports, so that I can share my `IoHandler<Metadata>` between transports.

tcp/src/line_codec.rs
NikVolf: yep, it's 0x0d0a sequence i think

tcp/src/meta.rs
NikVolf: Still unsure how it will work
But i can move it to stratum lib, k

NikVolf:
@tomusdrw issues addressed

PR #31, opened 2017-01-18
TCP migration to tokio, part II

NikVolf:
- dispatching messages for peers

### DIFF
--- a/tcp/src/dispatch.rs
+++ b/tcp/src/dispatch.rs
@@ -18,66 +18,65 @@ use std;
 use std::net::SocketAddr;
 use std::sync::{Arc, Mutex};
 
-use futures::{task, Stream, Poll, Async};
+use futures::{Stream, Poll, Async, Sink, Future};
+use futures::sync::mpsc;
 
-use std::collections::{HashMap, VecDeque};
-use std::collections::hash_map::Entry;
+use std::collections::HashMap;
 
-
-pub type MessageQueue = Mutex<HashMap<SocketAddr, VecDeque<String>>>;
-pub type TaskNotificationQueue = Mutex<Vec<task::Task>>;
+pub type SenderChannels = Mutex<HashMap<SocketAddr, mpsc::Sender<String>>>;
 
 pub struct PeerMessageQueue<S: Stream> {
     up: S,
-    queue: Arc<MessageQueue>,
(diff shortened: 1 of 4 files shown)


tcp/src/dispatch.rs
```diff
@@ -0,0 +1,125 @@
...
+use futures::{task, Stream, Poll, Async};
+
+use std::collections::{HashMap, VecDeque};
+use std::collections::hash_map::Entry;
+
+
+pub type MessageQueue = Mutex<HashMap<SocketAddr, VecDeque<String>>>;
+pub type TaskNotificationQueue = Mutex<Vec<task::Task>>;
+
+pub struct PeerMessageQueue<S: Stream> {
```
NikVolf: This stream produces same data as `up` stream + also adds messages from the queue

It parks current task when there are neither request responses nor queued messages and unparks once any message is pushed (task can also be unparked by the `up` stream itself - chacked in `message` test)

NikVolf:
@tomusdrw 
Sure it looks cleaner, will try to use this, thanx!

NikVolf:
Well seems that your suggestion won't work that well (at least i cannot make it so)

There is a problem with this sender lifetime

If we save it somewhere inside the server, then peer service will never end because receiver will never stop the stream (that's the receiver-sender channel behavior)

if we don't save, we won't be able to send anything to it other than via some polling queue which will again do tasks parking/unparking

which is already implemented here

so i can't see how mpsc channel can make things cleaner/easier here

tcp/src/tests.rs
```diff
@@ -171,3 +171,99 @@ fn peer_meta() {
...
+
+    thread::spawn(move || server.run().expect("Server must run with no issues"));
+    wait(100);
+
+    let mut core = Core::new().expect("Tokio Core should be created with no errors");
+    let timeout = Timeout::new(::std::time::Duration::from_millis(100), &core.handle())
+        .expect("There should be a timeout produced in message test");
+    let mut buffer = vec![0u8; 1024];
+    let mut buffer2 = vec![0u8; 1024];
+    let executed_dispatch = Mutex::new(false);
```
tomusdrw: Wouldn't `RefCell` be enough? Where is the `Sync` requirement comming from?

Issue #16, opened 2016-12-13
Use serde_derive instead of macros


tomusdrw:
https://users.rust-lang.org/t/serde-transitioning-to-macros-1-1/7437

PR #38, opened 2017-02-04
Update serde dependency to 0.9.0

Marwes:
This lets jsonrpc build on stable (1.15) without needing serde_codegen


tomusdrw:
Closes #16

core/src/lib.rs
```diff
@@ -22,12 +22,11 @@
 //! ```
 
 #![warn(missing_docs)]
-#![cfg_attr(feature="nightly", feature(custom_derive, plugin))]
-#![cfg_attr(feature="nightly", plugin(serde_macros))]
 
 #[macro_use] extern crate log;
 extern crate serde;
 extern crate serde_json;
+#[macro_use] extern crate serde_derive;
```
rphmeier: style: would prefer to keep `#[macro_use]` links adjacent

rphmeier:
tiny style nit but LGTM. Out of curiosity, how does this affect compile times?

Marwes:
> Out of curiosity, how does this affect compile times?

Running `time cargo test -p jsonrpc-core` gives the same times for me as before this change.

rphmeier:
All good then, thanks!

debris:
side note, we'll been quite substantial refactoring to switch `parity` to `serde 0.9`

Marwes:
When do you think you can release a version with this PR? Using a git dependency right now since I can't use serde 0.8.0 due to another dependency. It works for development but I can't do an actual release to crates.io with it.

PR #45, opened 2017-02-20
HTTP implementation using tokio-minihttp

### LINKED ISSUE #11: idea, use lightweight http server
Currently `jsonrpc-http-server` uses `hyper`. And since I remember there were some issues with it (that's why we have forked it). What do you think about replacing hyper with some other / smaller library? eg. [tiny-http](https://github.com/frewsxcv/tiny-http)

tomusdrw:
Closes #11


http/src/lib.rs
```diff
@@ -235,78 +182,177 @@ impl<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>> ServerBuilder<M, S> {
...
 	handle: Option<thread::JoinHandle<()>>,
-	event_loop: Option<RpcEventLoopHandle>,
+	close: Option<futures::sync::oneshot::Sender<()>>,
 }
 
 impl Server {
 	/// Returns addresses of this server
 	pub fn addrs(&self) -> &[SocketAddr] {
-		self.server.as_ref().unwrap().addrs()
+		&self.addr
```
rphmeier: It's no longer "addresses". Couldn't we just replace the `[SocketAddr; 1]` with a `SocketAddr` and return that? Is there a lot of API breakage?

tomusdrw: There are other breaking changes so it's fine.

http/src/lib.rs
```diff
@@ -235,78 +182,177 @@ impl<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>> ServerBuilder<M, S> {
...
+				meta_extractor: meta_extractor.clone(),
+				hosts: hosts.read().clone(),
+				cors_domains: cors_domains.clone(),
+			})).expect("Cannot bind to socket.");
+
+			let local_addr = server.local_addr().expect("OK");
+			// Add current host to allowed headers.
+			// NOTE: we need to use `local_address` instead of `addr`
+			// it might be different!
+			*hosts2.write() = Self::update_hosts(allowed_hosts, local_addr.clone());
```
maciejhirsz: Not sure what the intent here. Depending on the order of execution of the closure passed to `server.bind`, the `RpcService` will either get the updated or not updated hosts vec (I assume the closure is async).

Wouldn't it be easier to skip the Arc and RwLock altogether and update the list on stack before the closure?

tomusdrw: The problem is that we can only learn what address is we're listening on after we call `bind` which requires the closure.

There is no risk of race condition here, cause the server is just bound to an interface, but is not started yet, so the closure is never invoked before we update the `hosts`.

maciejhirsz: Maybe just obtain the write lock before the `server.bind` call, so it's more explicit that the code in the closure has to wait for the hosts to be updated?

tomusdrw:
Issues addressed.

NikVolf:
need merge

@rphmeier 
@debris 
ok to merge?

PR #48, opened 2017-02-25
Pub/Sub crate

### LINKED ISSUE #53: Missing docs for tcp crate

tomusdrw:
Initial implementation of publish subscribe. Example uses `jsonrpc-tcp-server`.

Outstanding issues:
- [ ] Implementation for IPC #50 
- [ ] Macros support #51 
- [ ] Websockets Server #49 

It will be implemented as separate PRs, will also create corresponding issues.

Closes #53


pubsub/src/subscription.rs
```diff
@@ -0,0 +1,389 @@
...
+use core::futures::sync::oneshot;
+
+use handler::{SubscribeRpcMethod, UnsubscribeRpcMethod};
+use types::{PubSubMetadata, SubscriptionId, TransportSender};
+
+/// RPC client session
+/// Keeps track of active subscriptions and unsubscribes from them upon dropping.
+pub struct Session {
+	active_subscriptions: Mutex<HashMap<(SubscriptionId, String), Box<Fn(SubscriptionId) + Send + 'static>>>,
+	transport: Mutex<TransportSender>,
```
NikVolf: Why do you need Mutex here?

tomusdrw: `TransportSender` is not `Send` and `Session` needs to be `Send`

NikVolf:
looks good to me!

PR #58, opened 2017-03-06
OPTIONS request 'Content-type' should not be json

dulanov:
Because of CORS browser send OPTIONS request and failed with '415 Unsupported Media Type'.


tomusdrw:
Looks good. Would be nice to include a test case for this (if you don't have time I'll prepare it tomorrow).

Also probably just altering the order of conditions should be sufficient (i.e. moving content type validation after `is_options` handling)

debris:
lgtm

PR #54, opened 2017-03-02
WebSockets server

tomusdrw:
Closes #49

PR #59, opened 2017-03-07
Shared server utils

### LINKED ISSUE #26: Abstract `RpcHandler` and `Remote` from `tokio-core`
`jsonrpc-core` should not have dependency on `tokio-core`.
`jsonrpc-core::reactor` could be moved to separate crate, but `RpcHandler` and `Remote` needs to be abstracted.

### LINKED ISSUE #55: Share `Remote` for all servers
Option to supply existing event loop remote instead of spawning one.

tomusdrw:
Contains event loop spawning/sharing, hosts validation and cors-headers.
Closes #55 
Closes #26 

Will need some additional work when #54 is merged.


http/src/handler.rs
```diff
@@ -142,26 +167,30 @@ impl<M: Metadata, S: Middleware<M>> server::Handler<HttpStream> for ServerHandle
...
-						.map_err(|e| format!("{:?}", e))
-						.and_then(|_| {
-							control.ready(Next::write()).map_err(|e| format!("{:?}", e))
-						});
-
-					if let Err(e) = result {
-						warn!("Error while resuming async call: {:?}", e);
-					}
+				self.jsonrpc_handler.remote.spawn(move |_| {
+						future.map(move |response| {
```
maciejhirsz: [code style] This block is indented deeper than it needs to be

tomusdrw: Good catch, cheers!

PR #60, opened 2017-03-07
Publish-Subscribe macros support

tomusdrw:
Closes: #51


tomusdrw:
roborob @rphmeier to the rescue! :1st_place_medal:

macros/src/auto_args.rs
```diff
@@ -76,21 +109,42 @@ macro_rules! build_rpc_trait {
...
+			)*
+
 		}
 	) => {
 		$(#[$t_attr])*
 		pub trait $name: Sized + Send + Sync + 'static {
-			/// Requests metadata
-			type Metadata: ::jsonrpc_core::Metadata;
+			metadata! (
+				$( $sub_name )*
```
rphmeier: a comment explaining that traits with subscriptions have different metadata than others might make sense :)

tomusdrw: Done.

rphmeier:
Looks good to me, but I think we're starting to reach a tipping point w.r.t. how much this macro can handle. If we need to add more features to it, IMO it would be better to take a different tack.

tomusdrw:
Agree. Submitted a separate issue: https://github.com/ethcore/jsonrpc/issues/66

PR #63, opened 2017-03-08
Middleware support for HTTP server

tomusdrw:


http/examples/http_middleware.rs
```diff
@@ -0,0 +1,29 @@
...
+
+fn main() {
+	let mut io = IoHandler::default();
+	io.add_async_method("say_hello", |_params| {
+		futures::finished(Value::String("hello".to_owned())).boxed()
+	});
+
+	let server = ServerBuilder::new(io)
+		.cors(DomainsValidation::AllowOnly(vec![AccessControlAllowOrigin::Null]))
+		.request_middleware(|request: &hyper::server::Request<hyper::net::HttpStream>| {
```
maciejhirsz: That was the thing where the compiler got confused about bound lifetimes without explicit type in the closure?

tomusdrw: Yes, but it's not the one that I was showing you the other day (that one was a metadata extractor while I was implementing `pubsub` crate, but the function type is the same).

PR #64, opened 2017-03-08
Pre processing CORS and Host headers 

### LINKED ISSUE #61: Pre-process hosts and cors headers
Currently you can supply any string to cors or hosts validation.
We do know however what format the strings should look like. Implement pre-procesing and validation for those.

tomusdrw:
Closes #61 

+ Version bump for all crates.


server-utils/src/hosts.rs
```diff
@@ -2,10 +2,76 @@
...
+		let host = hostname.next().unwrap();
+		let port = hostname.next().and_then(|port| port.parse().ok());
+
+		Host::new(host, port)
+	}
+
+	fn pre_process(host: &str) -> String {
+		// Remove possible protocol definition
+		let mut it = host.split("://");
+		let protocol = it.next().unwrap();
```
debris: maybe this method should return a `Result`?

server-utils/src/hosts.rs
```diff
@@ -2,10 +2,76 @@
...
+		let host = match it.next() {
+			Some(data) => data,
+			None => protocol,
+		};
+
+		let mut it = host.split('/');
+		it.next().unwrap().to_lowercase()
```
debris: same here

server-utils/src/hosts.rs
```diff
@@ -2,10 +2,76 @@
...
+		Host {
+			hostname: hostname,
+			port: port,
+			as_string: string,
+		}
+	}
+
+	/// Attempts to parse given string as a `Host`.
+	/// NOTE: This method always succeeds and falls back to sensible defaults.
+	pub fn parse(hostname: &str) -> Self {
```
debris: maybe `impl str::FromStr for Host` instead?

server-utils/src/hosts.rs
```diff
@@ -41,45 +107,58 @@ pub fn is_host_valid(host: Option<&str>, allowed_hosts: &Option<Vec<Host>>) -> b
...
-	let valid = is_host_valid(Some("any"), &None);
-	assert_eq!(valid, true);
-}
+	#[test]
+	fn should_parse_host() {
+		assert_eq!(Host::parse("http://parity.io"), Host::new("parity.io", None));
+		assert_eq!(Host::parse("https://parity.io:8443"), Host::new("parity.io", Some(8443)));
+		assert_eq!(Host::parse("chrome-extension://124.0.0.1"), Host::new("124.0.0.1", None));
+		assert_eq!(Host::parse("parity.io/somepath"), Host::new("parity.io", None));
+		assert_eq!(Host::parse("127.0.0.1:8545/somepath"), Host::new("127.0.0.1", Some(8545)));
```
debris: there are no tests for invalid host name

server-utils/src/cors.rs
```diff
@@ -1,9 +1,94 @@
...
+		let mut proto = it.next();
+		let mut hostname = it.next();
+
+		if hostname.is_none() {
+			hostname = proto;
+			proto = None;
+		}
+
+		let proto = proto.map(|s| s.to_lowercase());
+		let hostname = Host::parse(hostname.unwrap());
```
debris: it would be better to use `expect` here or even to refactor to:

```rust
let mut it = data.split("://").rev();
let mut hostname = Host::parse(it.next().expect("it.len() > 0; qed"));
let mut proto = it.next().map(String::to_lowercase);
```

server-utils/src/hosts.rs
tomusdrw: That line cannot fail. Any string will result in non-empty iterator when `split` is called. I will replace `unwraps` with `expect` though.

server-utils/src/hosts.rs
tomusdrw: That's the whole point.
I was considering implementation of `FromStr` instead, but then thought that it doesn't make much sense. I'm not really trying to validate if supplied hosts are actually a valid addresses (that would be much harder), but rather trying to provide a convenience method to input them, pre-processing invalid entries.

debris:
👍

PR #65, opened 2017-03-09
Shared remote for websockets server.

tomusdrw:


ws/src/server.rs
```diff
@@ -76,14 +85,16 @@ impl Server {
 	}
 
 	/// Closes the server and waits for it to finish
-	pub fn close(self) {
+	pub fn close(mut self) {
 		let _ = self.broadcaster.shutdown();
+		self.remote.take().unwrap().close();
```
maciejhirsz: `self.remote().take().map(|remote| remote.close());` to avoid unwrap?

tomusdrw: It should actually be `expect`, we can prove that remote is always there. Will fix.

tomusdrw: Fixed.

PR #69, opened 2017-03-11
Reimplement IPC with tokio_uds

NikVolf:


ipc/Cargo.toml
```diff
@@ -17,6 +17,10 @@ bytes  = "0.3.0"
 miow = "0.1"
 log = "0.3"
 rand = "0.3"
+tokio-core = "0.1"
```
tomusdrw: You can reference `tokio-core` from `server-utils`, will simplify migrations to newer versions (less changes in dependent crates)
Same for `futures`, it's re-exported by `jsonrpc-core`

ipc/src/stream_codec.rs
```diff
@@ -0,0 +1,137 @@
+extern crate tokio_core;
+
+use std::io;
+use stream_codec::tokio_core::io::{Codec, EasyBuf};
+
+pub struct StreamCodec;
+
+fn is_whitespace(byte: u8) -> bool {
+	match byte {
+		0x0D | 0x0A | 0x20 | 0x09 => true,
```
tomusdrw: Spaces (`0x20`) and Tabs (`0x09`) can occur in JSON strings, so doesn't it cause some issues with the condition in L48?

ipc/src/stream_codec.rs
```diff
@@ -0,0 +1,137 @@
...
+			let byte = buf.as_ref()[idx];
+
+			if (byte == b'{' || byte == b'[') && !in_str {
+				if depth == 0 {
+					start_idx = idx;
+				}
+				depth += 1;
+			}
+			else if (byte == b'}' || byte == b']') && !in_str {
+				depth -= 1;
```
tomusdrw: Will overflow in case `}}` is received.

ipc/src/lib.rs
```diff
@@ -33,15 +34,20 @@ extern crate mio;
 #[cfg(not(windows))]
 extern crate bytes;
```
tomusdrw: Some crates are also imported directly in `ipc/src/uds.rs`, might be good to keep everything in a single place.

ipc/src/uds.rs
```diff
@@ -0,0 +1,287 @@
...
+				let service = Service::new(rpc_handler.clone(), meta);
+				let (writer, reader) = unix_stream.framed(StreamCodec).split();
+				let responses = reader.and_then(
+					move |req| service.call(req).then(|response| match response {
+						Err(e) => {
+							warn!(target: "uds", "Error while processing request: {:?}", e);
+							future::ok(String::new())
+						},
+						Ok(None) => {
+							trace!(target: "uds", "JSON RPC request produced no response");
```
tomusdrw: It's a valid behaviour when notification is received (no `id`) - server should not respond anything in such case.

ipc/src/uds.rs
```diff
@@ -0,0 +1,287 @@
...
+	remote: Option<reactor::Remote>,
+	stop: Option<oneshot::Sender<()>>,
+}
+
+impl Server {
+	/// Closes the server (waits for finish)
+	pub fn close(mut self) {
+		self.stop.take().unwrap().complete(());
+		self.remote.take().unwrap().close();
+		self.clear_file();
```
tomusdrw: Shouldn't it also wait for finish as docs state?

ipc/src/uds.rs
```diff
@@ -0,0 +1,287 @@
...
+		io.add_method("say_hello", |_params| {
+			Ok(Value::String("hello".to_string()))
+		});
+		ServerBuilder::new(io)
+	}
+
+	fn run(path: &str) -> Server {
+		let builder = server_builder();
+		let server = builder.start(path).expect("Server must run with no issues");
+		thread::sleep(::std::time::Duration::from_millis(50));
```
tomusdrw: Is this `sleep` needed? Don't we wait for the server to start anyway?

ipc/src/win.rs
```diff
@@ -292,3 +292,16 @@ impl<M: Metadata, S: Middleware<M> + 'static> Drop for Server<M, S> {
 		trace!(target: "ipc", "IPC Server : shutdown");
 	}
 }
+
+
+pub fn server<I, M: Metadata, S: Middleware<M> + Send + Sync + 'static>(
```
tomusdrw: `Send + Sync + 'static` bounds are not necessary here, they are part of `Middleware` trait: https://github.com/ethcore/jsonrpc/blob/b36a57c4bf449c431dc59f0e88236026eda62ea7/core/src/middleware.rs#L8

ipc/src/uds.rs
```diff
@@ -0,0 +1,287 @@
...
+
+impl Drop for Server {
+	fn drop(&mut self) {
+		self.stop.take().map(|stop| stop.complete(()));
+		self.remote.take().map(|remote| remote.close());	
+		self.clear_file();	
+	}
+}
+
+pub fn server<I, M: Metadata, S: Middleware<M> + Send + Sync + 'static>(
```
tomusdrw: `Send + Sync + 'static` bounds are not necessary here, they are part of `Middleware` trait: https://github.com/ethcore/jsonrpc/blob/b36a57c4bf449c431dc59f0e88236026eda62ea7/core/src/middleware.rs#L8

ipc/src/stream_codec.rs
NikVolf: It's ok, it will just ignore requests that consists only from this characters (without even '{' and '"')

ipc/src/uds.rs
NikVolf: No, we don't wait, it's needed for tests only

ipc/src/win.rs
NikVolf: Why have you added it to the TCP then? :)

tomusdrw: That's a very good question:) No idea :)

ipc/src/stream_codec.rs
NikVolf: Nope, i guess default integer type is i32

ipc/src/lib.rs
NikVolf: removed several unused externs

ipc/src/uds.rs
NikVolf: isn't remote.close() joins the spawned thread?

NikVolf:
@tomusdrw 
All addressed / answered

ipc/src/stream_codec.rs
NikVolf: so we don't trim anything, we just don't consider some requests as a compete ones until there is at least one non-whitespace character in it

PR #73, opened 2017-03-15
Don't process invalid CORS requests

### LINKED ISSUE #62: Don't process requests when invalid cross-origin request is received
Currently http and minihttp servers are processing the request even if `Origin` header is not allowed to access. The requester will not be able to read the response, but can cause side effects to happen on the node.
We should rather fail fast in such case and avoid even processing the request.

tomusdrw:
Closes #62 

Also returns correct headers for OPTIONS requests.

### DIFF
--- a/http/src/request_response.rs
+++ b/http/src/request_response.rs
@@ -60,7 +60,7 @@ impl Response {
 	pub fn host_not_allowed() -> Self {
 		Response {
 			code: StatusCode::Forbidden,
-			content_type: header::ContentType::html(),
+			content_type: header::ContentType::plaintext(),
 			content: "Provided Host header is not whitelisted.\n".to_owned(),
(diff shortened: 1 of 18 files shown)


http/src/request_response.rs
```diff
@@ -81,6 +85,16 @@ impl Response {
 			write_pos: 0,
 		}
 	}
+
+	/// CORS invalid
+	pub fn invalid_cors() -> Self {
+		Response {
+			code: StatusCode::Forbidden,
+			content_type: header::ContentType::html(),
```
maciejhirsz: `ContentType::plaintext()` would match `invalid_cors` from minihttp.

tomusdrw:
Beta and Nightly failures are not related no this PR and look like compiler regressions.

PR #81, opened 2017-03-19
use hyper from cargo (in progress)

debris:


http/src/handler.rs
```diff
@@ -1,292 +1,145 @@
...
+					Method::Post if is_json(request.headers.get::<header::ContentType>()) => {
+						// TODO: handle jsonrpc request here
+						//read_to_end(RequestRead(&request), Vec::new())
+							//.map_err(|_err| Error::Other)
+							//.and_then(move |(request, read)| {
+								//Ok(String::new())
+							//}).boxed()
+							////.and_then(
+						//Ok(String::new())
+						future::ok(String::new()).boxed()
```
debris: @tomusdrw I dunno how to make this part work.

debris:
I've tried to use lastest `hyper` from cargo - `0.10.5`. Don't know if it's worth to carry on with this pr now. Might be worth waiting for `0.11` with integrates `futures` library

http/src/lib.rs
```diff
@@ -254,76 +236,22 @@ impl<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>> ServerBuilder<M, S> {
...
-			server: Some(l),
-			handle: Some(handle),
-			remote: Some(eloop),
+			server: Some(listening),
 		})
 	}
 }
 
 /// jsonrpc http server instance
 pub struct Server {
```
tomusdrw: Server methods were useful:
1. `addrs` - getting actuall address the server is listening on when using port `0` (random port)
2. `close` - programatically closing the server and waiting for it to finish (that could be replaced with `drop`, but I think it was nice to have it explicitly (also on drop we don't need to wait for completion, just sending close signal))
3. `wait` - used in examples or when you want to block a thread and wait for the server to finish (without sending a close signal)

http/src/lib.rs
```diff
@@ -254,76 +236,22 @@ impl<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>> ServerBuilder<M, S> {
...
-		let handle = thread::spawn(move || {
-			srv.run();
-		});
+		let listening = hyper::Server::http(addr)?.handle(server_handler)?;
 
 		Ok(Server {
-			server: Some(l),
-			handle: Some(handle),
-			remote: Some(eloop),
+			server: Some(listening),
```
tomusdrw: EventLoop handle missing (it will be closed immediately when dropped)

http/src/lib.rs
```diff
@@ -109,40 +100,39 @@ impl<T: hyper::server::Handler<hyper::net::HttpStream> + Send + 'static> From<Op
...
 		Default::default()
 	}
 }
 
 impl<M, F> HttpMetaExtractor<M> for F where
 	M: jsonrpc::Metadata,
-	F: Fn(&server::Request<hyper::net::HttpStream>) -> M + Sync + Send + 'static,
-{
-	fn read_metadata(&self, req: &server::Request<hyper::net::HttpStream>) -> M {
+	F: Fn(&server::Request) -> M + Sync + Send + 'static {
```
tomusdrw: Formatting was better before :)

1. No special treatment of last line
2. Visual whitespace between bounds and beginning of implementation

http/src/lib.rs
```diff
@@ -109,40 +100,39 @@ impl<T: hyper::server::Handler<hyper::net::HttpStream> + Send + 'static> From<Op
 /// Allows to intercept request and handle it differently.
 pub trait RequestMiddleware: Send + Sync + 'static {
```
tomusdrw: Output future should be associated type.

http/src/handler.rs
tomusdrw: Hmm, wouldn't just:
```rust
let mut string = String::new();
request.read_to_string(&mut string).unwrap();

self.io.handle_request(&string, metadata).then(|response| {
  let string = response.unwrap_or("".to_owned());
  res.write(&string)
}).boxed()
```
work (or something like this)?

http/src/lib.rs
```diff
@@ -254,76 +236,22 @@ impl<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>> ServerBuilder<M, S> {
...
-				for addr in l.addrs() {
-					let address = addr.to_string();
-					new_hosts.insert(address.clone().into());
-					new_hosts.insert(address.replace("127.0.0.1", "localhost").into());
-				}
-				// Override hosts
-				*hosts = Some(new_hosts.into_iter().collect());
-			}
-		}
+		let server_handler = ServerHandler {
```
tomusdrw: Taking `address` from running instance was quite useful if someone started the server with `port=0` - it causes random port to be allocated and that's why we were updating the allowed hosts header and storing it in the handle.

http/src/lib.rs
```diff
@@ -109,40 +100,39 @@ impl<T: hyper::server::Handler<hyper::net::HttpStream> + Send + 'static> From<Op
...
 	/// Takes a request and decides how to proceed with it.
-	fn on_request(&self, request: &server::Request<hyper::net::HttpStream>) -> RequestMiddlewareAction;
+	fn on_request(&self, request: &server::Request) -> BoxFuture<RequestMiddlewareAction, ()>;
 }
 
 impl<F> RequestMiddleware for F where
-	F: Fn(&server::Request<hyper::net::HttpStream>) -> RequestMiddlewareAction + Sync + Send + 'static,
-{
-	fn on_request(&self, request: &server::Request<hyper::net::HttpStream>) -> RequestMiddlewareAction {
+	F: Fn(&server::Request) -> BoxFuture<RequestMiddlewareAction, ()> + Send + Sync + 'static {
```
tomusdrw: Formatting (as below)

http/src/lib.rs
debris: I'm afraid those methods do not exist any more. There were some many api changes in hyper, that instead of removing the code line by line, I just started from scratch and begin to pull your code there.

Nevertheless, if I continue with this pr, I will try to address these problems ;)

debris:
closed in favor of #101

PR #94, opened 2017-03-30
Transport dropping

rphmeier:
Drop requests issued within a `ws::Session` when that session is dropped.


rphmeier:
I can't figure out the test failure here, it's not reproducible locally. @tomusdrw do you have any suggestion on a non racy way to see if the future actually gets dropped from the event loop?

rphmeier:
Ok, seems that hyper isn't calling `Handler::on_remove` on some machines.

rphmeier:
(will wait on #101 for new http)

tomusdrw:
@rphmeier Mind updating the PR? I think only WS server requires explicit dropping now (the rest of the transports is future-based).

rphmeier:
@tomusdrw both `HEAD` and `HEAD~1` are working implementations. The more recent commit eagerly allocates a `oneshot` channel on the `Slab` for each RPC call, but is much simpler code. The commit before that used a scheme with one side of a `futures::sync::BiLock` used for blocking access (`LivenessPoll` and `Session` drop) and the other side used for async access (only used to lazily allocate the channel on first poll). 

If it seems to be a bottleneck we can try the lazy version as well.

ws/src/session.rs
```diff
@@ -72,6 +77,61 @@ impl From<Option<ws::Response>> for MiddlewareAction {
...
+impl LivenessPoll {
+	fn create(task_slab: Arc<TaskSlab>) -> Self {
+		const INITIAL_SIZE: usize = 4;
+
+		let (index, rx) = {
+			let mut task_slab = task_slab.lock();
+			if !task_slab.has_available() {
+				// grow the size if necessary.
+				// we don't expect this to get so big as to overflow.
+				let reserve = ::std::cmp::max(task_slab.capacity(), INITIAL_SIZE);
```
tomusdrw: Will it ever grow?

tomusdrw:
I think we'll be ok with eager allocation, it won't be needed when #158 is implemented (hopefuly for `8.0`)

ws/src/session.rs
rphmeier: yep, it'll grow up to `next_power_of_two(maximum amount of live requests in session)`

rphmeier: it starts at zero though, so it only lazily allocates the slab memory when the first request comes in, and odds are there will rarely be more than 4-8 requests at a given time

PR #95, opened 2017-03-31
Parity-related improvements

tomusdrw:
- Unify http and minihttp APIs
- Add `control` to HttpMiddleware


http/src/lib.rs
```diff
@@ -55,21 +55,21 @@ pub type ServerResult = Result<Server, Error>;
 #[derive(Debug)]
 pub enum Error {
 	/// IO Error
-	IoError(std::io::Error),
+	Io(std::io::Error),
```
maciejhirsz: :+1:

Issue #91, opened 2017-03-29
Drop request futures in transports upon hangup


rphmeier:

tomusdrw:
Currently we just submit the future returned by `jsonrpc-core` directly to the event loop. We should wrap it in a transport specific future that will drop upstream future when connection/request is dropped.

tomusdrw:
Closed in #94

PR #101, opened 2017-04-04
Use latest hyper (master)

tomusdrw:
Implements `jsonrpc-http-sever` using future-based `master` branch of `hyper`.
Related #91 

Also fixes couple of TODOs (unecessary Boxing in minihttp and `io::Error`s propagation)

Performance stats (example `server`, compiled in `--release`)
1. Old http: 2.157 req/s
2. New http: 23.456 req/s
3. Mini http (threads=1): 33.822 req/s


http/src/handler.rs
```diff
@@ -70,87 +63,227 @@ impl<M: Metadata, S: Middleware<M>> server::Handler<HttpStream> for ServerHandle
...
+	is_options: bool,
+	cors_header: cors::CorsHeader<header::AccessControlAllowOrigin>,
+}
+
+impl<M: Metadata, S: Middleware<M>> Future for RpcHandler<M, S> {
+	type Item = server::Response;
+	type Error = hyper::Error;
+
+	fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+		let new_state = match mem::replace(&mut self.state, RpcHandlerState::Done) {
```
debris: why is this `mem::replace` needed? imo, it would be cleaner to set state only in one place (at the end), with simple assign

http/src/handler.rs
```diff
@@ -70,87 +63,227 @@ impl<M: Metadata, S: Middleware<M>> server::Handler<HttpStream> for ServerHandle
...
+					Ok(Async::Ready(response)) => {
+						RpcPollState::Ready(RpcHandlerState::Writing(match response {
+							// Notification, just return empty response.
+							None => Response::ok(String::new()),
+							// Add new line to have nice output when using CLI clients (curl)
+							Some(result) => Response::ok(format!("{}\n", result)),
+						}.into()))
+					},
+					Ok(Async::NotReady) => RpcPollState::NotReady(RpcHandlerState::Waiting(waiting)),
+					Err(_) => RpcPollState::Ready(RpcHandlerState::Writing(Response::internal_error())),
```
debris: is it possible to write response at this point? If it is `hyper::Error` we are most likely disconnected from the node, aren't we?

http/src/handler.rs
```diff
@@ -70,87 +63,227 @@ impl<M: Metadata, S: Middleware<M>> server::Handler<HttpStream> for ServerHandle
...
+	fn read_headers(
+		&self,
+		request: server::Request,
+		continue_on_invalid_cors: bool,
+	) -> RpcHandlerState<M> {
+		if self.cors_header == cors::CorsHeader::Invalid && !continue_on_invalid_cors {
+			return RpcHandlerState::Writing(Response::invalid_cors());
+		}
+		// Read metadata
+		let metadata = self.jsonrpc_handler.extractor.read_metadata(&request);
```
debris: can metadata be location in `request.body()`, if yes, maybe extractor should return the future? or maybe we should read `request.body()` stream, before calling this function?

http/src/handler.rs
```diff
@@ -70,87 +63,227 @@ impl<M: Metadata, S: Middleware<M>> server::Handler<HttpStream> for ServerHandle
...
+					let content = match ::std::str::from_utf8(&request) {
+						Ok(content) => content,
+						Err(_) => {
+							// returns empty response on invalid utf8
+							return Ok(RpcPollState::Ready(RpcHandlerState::Writing(Response::empty())));
+						},
+					};
+
+					return Ok(RpcPollState::Ready(RpcHandlerState::Waiting(
+						self.jsonrpc_handler.handler.handle_request(content, metadata)
```
debris: this function is called `read_body`. imo it should not handle the request, just read the body ;)

http/src/handler.rs
```diff
@@ -70,87 +63,227 @@ impl<M: Metadata, S: Middleware<M>> server::Handler<HttpStream> for ServerHandle
...
+
+	fn read_body(
+		&self,
+		mut body: hyper::Body,
+		mut request: Vec<u8>,
+		metadata: M,
+	) -> Result<RpcPollState<M>, hyper::Error> {
+		loop {
+			match body.poll()? {
+				// TODO [ToDr] reject too large requests?
```
debris: maybe use `tokio_io::io::read_all` or `tokio_io::io::read_exact` instead?

http/src/handler.rs
```diff
@@ -70,87 +63,227 @@ impl<M: Metadata, S: Middleware<M>> server::Handler<HttpStream> for ServerHandle
...
+				Async::Ready(Some(chunk)) => {
+					request.extend_from_slice(&*chunk)
+				},
+				Async::Ready(None) => {
+					let content = match ::std::str::from_utf8(&request) {
+						Ok(content) => content,
+						Err(_) => {
+							// returns empty response on invalid utf8
+							return Ok(RpcPollState::Ready(RpcHandlerState::Writing(Response::empty())));
```
debris: it should be a proper error message instead

http/src/handler.rs
tomusdrw: We want to take the ownership of some stuff stored in the current state (like `ReadingHeaders { request, cors_domains, ...}`).

http/src/handler.rs
tomusdrw: It's not really `hyper::Error` that we are getting here waiting is `BoxFuture<Option<String>, ()>` and it's an error coming from `jsonrpc_core` which (in theory) can never fail. That's why we return `internal_error` here.

http/src/handler.rs
tomusdrw: Not really, metadata is only stored in Headers, `body` is the request/data itself. I don't think we should support that, cause if request contains more data then specified by JSON-RPC protocol then metadata extractor should actually somehow strip this data before processing it further which makes it more like a `jsonrpc_core::Middleware` then transport-specific metadata extractor.

http/src/handler.rs
tomusdrw: Currently `hyper::Body` doesn't implement `AsyncRead` trait required by those functions. It's a `Stream` of `Chunk`s that happen to have `Deref<Target=[u8]>`. Maybe it'll be changed in future in hyper.

http/src/handler.rs
debris: so maybe only headers should be passed as a param?

http/src/handler.rs
debris: > We want to take the ownership of some stuff stored in the current state (like ReadingHeaders { request, cors_domains, ...}).

I see 👍

http/src/handler.rs
tomusdrw: Headers are parsed on demand, you would need to know in advance what headers are required by specific extractor (possible but IMHO unnecessary).

http/src/handler.rs
debris: 😞

http/src/handler.rs
tomusdrw: Ah, sorry I see your point  - just to pass `Headers` struct ref? Yeah we could do that by I can imagine using `path()/query()` as part of metadata as well.

http/src/handler.rs
rphmeier: You could potentially write a wrapper around it for `AsyncRead`, but it would look basically the same as this code.

tomusdrw:
@debris issues addressed.

rphmeier:
Multithreading it would be cool!

tomusdrw:
Yup and afaict it should be possible (even easy). Logged here: https://github.com/paritytech/jsonrpc/issues/102

PR #103, opened 2017-04-05
Support multiple hyper threads.

### LINKED ISSUE #102: Multithreaded hyper server
Should attempt to re-use port on *nix, like here:
https://tokio-rs.github.io/tokio-proto/src/tokio_proto/tcp_server.rs.html#212

tomusdrw:
Closes #102 

Benchmarks:
- 1 thread: 21,374 req/s
- 4 threads 75,751 req/s

### DIFF
--- a/http/Cargo.toml
+++ b/http/Cargo.toml
@@ -11,6 +11,7 @@ documentation = "https://paritytech.github.io/jsonrpc/jsonrpc_http_server/index.
 
 [dependencies]
 log = "0.3"
+net2 = "0.2"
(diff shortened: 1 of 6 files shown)


rphmeier:
Not a huge fan of the re-binding of the TCP listener, but there isn't really another way yet. When tokio changes their API to allow listening on multiple threads more cleanly I think we should port it over.

tomusdrw:
@debris care to have a look?

debris:
lgtm, I reviewed that 6 days ago, just forgot to press `accept` 😛

PR #110, opened 2017-04-12
Additional improvements to integrate pubsub with Parity

### LINKED ISSUE #109: Implement mising futures::Sink implementations
`pubsub::Sink` and `macros::pubsub::Sink` should implement `futures::Sink`.

Actually instead of using those structures it should be possible to do `self.transport.with(Box::new(FnMut))`  and just declare those structs as type aliases to `futures::sink::With`.

tomusdrw:
Closes #109

### DIFF
--- a/macros/src/pubsub.rs
+++ b/macros/src/pubsub.rs
@@ -5,6 +5,8 @@ use jsonrpc_pubsub as pubsub;
 use serde;
 use util::to_value;
 
+use self::core::futures::{self, Sink as FuturesSink};
+
 pub use self::pubsub::SubscriptionId;
 
 pub struct Subscriber<T> {
@@ -29,6 +31,7 @@ impl<T> Subscriber<T> {
 		Ok(Sink {
 			id: id,
 			sink: sink,
+			buffered: None,
 			_data: PhantomData,
 		})
 	}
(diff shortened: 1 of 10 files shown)


macros/src/pubsub.rs
```diff
@@ -37,16 +40,64 @@ impl<T> Subscriber<T> {
...
+	}
+}
+
+impl<T: serde::Serialize> futures::sink::Sink for Sink<T> {
+	type SinkItem = T;
+	type SinkError = pubsub::TransportError;
+
+	fn start_send(&mut self, item: Self::SinkItem) -> futures::StartSend<Self::SinkItem, Self::SinkError> {
+		if self.poll()?.is_not_ready() {
+			return Ok(futures::AsyncSink::NotReady(item));
```
rphmeier: does this need to schedule a `Task` wakeup?

rphmeier: I'm pretty sure it doesn't, but putting the rationale why in a comment above (since futures have a lot of hidden complexity) would help

debris:
I am not familiar with how `future::Sink` works, so I can only assume that the logic is 100% correct. The pr itself looks good

macros/src/pubsub.rs
tomusdrw: Updated!

tomusdrw:
Nightly failure seems unrelated to the changes. See https://github.com/rust-lang-nursery/rustup.rs/issues/1062 for more details.

PR #115, opened 2017-04-18
Windows IPC implementation fails in case of multimple concurrent requests

kirushik:
This is just a PR to start a discussion.

I'd really like to improve windows support for the IPC implementation.
The usecase I'm facing is the following: multiple clients are asking server for a long and heavy computation simultaneously.

Currently the linux sockets implementation sorta-works (parallel processing aside), while windows pipes-based one deadlocks somehow (if I'm using synchronous API for handlers, `add_method`), or fails with `Error { repr: Os { code: 5, message: "Access is denied." } }` (in case of future-based `add_async_method` implementation).

Either I am doing something really wrong here (it's quite possible, I'm not a Windows guy; will be grateful for any tips), or Windows implementation needs some love (and then I'd like to contribute, but might need some directions on from where to start; still not a Windows guy).

Please feel free to close this PR if needed -- after all my tests currently won't work on `master`, only on `stable`.

### DIFF
--- a/ipc/tests/multithreaded_test.rs
+++ b/ipc/tests/multithreaded_test.rs
@@ -1,0 +1,59 @@
+extern crate jsonrpc_core;
+extern crate jsonrpc_ipc_server;
+
+extern crate miow;
+
+#[cfg(test)]
+mod multithreaded_test {
+	use jsonrpc_core::*;
+	use jsonrpc_ipc_server::Server;
+
+	use miow::pipe::connect;
+
+	use std::thread;
+	use std::time::Duration;
+	use std::io::{Read, Write};
+
+	#[cfg(windows)]
+	fn pipe_name() -> &'static str {
+		"\\\\.\\pipe\\Foo\\Bar\\Baz"
+	}
+
+	#[cfg(windows)]
+	fn say_to_pipe(pipe_name: &str, message: String) -> String {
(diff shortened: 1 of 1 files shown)


ipc/tests/multithreaded_test.rs
```diff
@@ -0,0 +1,106 @@
...
+        thread::spawn(move || server.run());
+
+        thread::sleep(Duration::from_millis(100)); // Let's make sure the pipe server has been initialized
+
+        let thread1 = thread::spawn(|| {
+                                        assert_eq!(say_to_pipe(pipe_name(), message(1)),
+                                                   expected_response(1))
+                                    });
+        let thread2 = thread::spawn(|| {
+                                        assert_eq!(say_to_pipe(pipe_name(), message(2)),
```
NikVolf: the problem is not in the multiple connections, but in simultanious ones
if you add 1ms delay before that line, the problem will dissapear

of course, this is still a bug

i thought the problem might be addressed by this pr
https://github.com/paritytech/jsonrpc/pull/77

will try it tomorrow

NikVolf:
@kirushik 

so i included your test and fix for it here:
https://github.com/paritytech/jsonrpc/pull/116

care to test on your setup?

kirushik:
Superceded by #116 , closing this one.

PR #116, opened 2017-04-19
kirushik tests and fixes using async connect for stable branch

NikVolf:


ipc/tests/multithreaded_test.rs
```diff
@@ -0,0 +1,133 @@
...
+#[cfg(not(windows))]
+fn pipe_name() -> &'static str {
+    "/tmp/foobar.sock"
+}
+
+#[cfg(windows)]
+fn say_to_pipe(pipe_name: &str, message: String) -> String {
+    let mut connection =
+        connect(pipe_name).expect("Failed to get a client connection to the pipe");
+    connection
```
tomusdrw: Seems that only the first line could be platform specific, the rest of the code is duplicated.

ipc/src/win.rs
```diff
@@ -85,38 +85,47 @@ impl<M: Metadata, S: Middleware<M>> PipeHandler<M, S> {
...
-			try!(NamedPipeBuilder::new(addr)
-				.first(false)
-				.accept_remote(true)
-				.inbound(true)
-				.outbound(true)
-				.out_buffer_size(MAX_REQUEST_LEN)
-				.in_buffer_size(MAX_REQUEST_LEN)
-				.create()));
+		let mut connected_pipe = {
+			let mut pipe = self.waiting_pipe.lock().unwrap();
```
tomusdrw: Why do we keep the pipe locked the whole time? Wouldn't it be more sensible to release the lock when we're waiting?

kirushik: The following also passes the test for me:
```rust
loop {
	if {
		let pipe = self.waiting_pipe.lock().unwrap();
		unsafe {
			! try!(pipe.connect_overlapped(overlapped.raw()))
		}
	} {
```
and then several lines below, 
```rust
let mut pipe = self.waiting_pipe.lock().unwrap();
std::mem::replace::<NamedPipe>(&mut *pipe,
```

Clearly mutexes are more short-lived and granular this way, but maybe it has some consequences  for Windows pipe handling somehow?

kirushik:
@NikVolf Yes, this fixes my test, awesome!

NikVolf:
@kirushik thanx for everything!

kirushik:
@NikVolf No problems at all. Won't be the last contribution here, if my idea sticks...

Are you going to make a bugfix release 6.0.1 of a `jsonrcp-ipc-server` on crates.io for this?
`git` cannot be used together with `path` for dependencies in `Cargo.toml`, AFAIK.

NikVolf:
@kirushik i don't think so
the stable branch is supported for backward compatibility only, active branch is master and there is completely different transport and api there for uds/pipes.

danielkornev:
@NikVolf does it mean you have a v7 of a ```jsonrpc-ipc-server``` being prepared for eventual release on crates.io?

kirushik:
@NikVolf Can't say I completely understood your decision (I thought fixed current version would be `6.0.1`, and your new one from master should be `7.0.0` since it introduces breaking changes...) — but that's certainly up to you to decide how to maintain your releases.

Do you mind if I'll do some `git filter-branch` magic and rehost the ipc subfolder under my account?
Not only LICENSE, but even authorship data (from commits history) will be preserved, of course.

NikVolf:
sure, there will be 7.* versions on crates.io eventually, but i don't think we will be publishing 6.* versions anymore

if you need v6 with this fix just reference to git directly

@kirushik np

PR #117, opened 2017-04-19
Implement the Error trait on ws Error type

faern:
Implement the Error trait on the websocket server error type. Same reason as in #97

### DIFF
--- a/ws/src/server_builder.rs
+++ b/ws/src/server_builder.rs
@@ -1,3 +1,4 @@
+use std::fmt;
 use std::io;
 use std::net::SocketAddr;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


ws/src/server_builder.rs
```diff
@@ -37,6 +38,28 @@ impl From<io::Error> for Error {
...
+        match *self {
+            Error::Io(ref e) => e.fmt(f),
+            Error::WebSocket(ref e) => e.fmt(f),
+        }
+    }
+}
+
+impl ::std::error::Error for Error {
+    fn description(&self) -> &str {
+        "Starting the JSON-RPC WebSocket server failed"
```
tomusdrw: Why not delegate to wrapped errors?

faern: I wanted them to work the same as the http errors. Also a chain of errors (this is one error and it contains a cause that is another error) should not just chain down transparently. The intermediate steps in the chain must have relevant information, otherwise there is no information why one got the bottom level error one got.

Instead of just delegating this down one might even want to change the display implementation to be `"Starting the JSON-RPC WebSocket server failed: {}"` and then format in the cause.

faern: I have been looking in the rustc code for any standard on how the `description` and `Display` are implemented. But I could not find a common denominator for enums containing sub-errors.

tomusdrw: Fair enough! :+1:

tomusdrw:
Could you merge with latest master so that we can see that the tests are not failing?

faern:
@tomusdrw It is based on the latest master? Or should I merge it with master on my fork, what does that help with?

faern:
@tomusdrw Latest master fails with the same error on my machine as well.

tomusdrw:
Oh, ok sorry. It's probably caused by a `Cargo.lock` file locally, for travis is must be cache. Will re-run the builds.

tomusdrw:
Sorry once again, actually I thought that the fix is already merged, but I just made a PR: #119 
You can wait for that one to be merged or just replace `set_body(b"Hello World") -> set_body("Hello World".to_owned())`.

faern:
@tomusdrw Fixed. Should hopefully build now.

faern:
Now nightly crashes on something else. But I get that locally as well. It's a bug that I think is reported already.

tomusdrw:
Yeah, the crash is irrelevant. Thank you so much for the PR!

Issue #120, opened 2017-04-20
Example/doc for how to talk to a PubSub server


faern:
I don't find very much information how to talk to a server using the PubSub feature of this repository. Would it be possible to add an example or some docs on that?

I have followed the example file to set up a server, and it is running. But I don't know what send to it to register my client.

Some help with this would be awesome. Thanks.

tomusdrw:
If you have a running server from this example:
https://github.com/paritytech/jsonrpc/blob/a95e135017c3a29fbda5dad88dc62f8797792443/pubsub/examples/pubsub.rs

You need to issue a standard RPC with `{"id":1,"jsonrpc":"2.0","method":"subscribe_hello"}` and then you should start seeing subscriptions coming to your client.
The transport used in the example is TCP, so you can for instance use `netcat` to connect to the server:
```
$ netcat localhost 3030 -
{"id":1,"jsonrpc":"2.0","method":"subscribe_hello"}
{"jsonrpc":"2.0","result":5,"id":1}
{"jsonrpc":"2.0","method":"hello","params":[10]}
{"jsonrpc":"2.0","method":"hello","params":[10]}
{"jsonrpc":"2.0","method":"hello","params":[10]}
```

faern:
I'm running the example using macros at https://github.com/paritytech/jsonrpc/blob/a95e135017c3a29fbda5dad88dc62f8797792443/macros/examples/pubsub-macros.rs

But I get this:
```
{"id":1,"jsonrpc":"2.0","method":"hello_subscribe"}
{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found"},"id":1}
```

faern:
I did get your example to work with the example not using macros. But using the macros would be sweet.

tomusdrw:
Hmm, you're right. It turned out that the alias invocation in `macros` was invalid (it was overriding the implementation with alias to non-existent method).

Please try with #121 or remove `alias` in the example.

faern:
That worked slightly better! Thanks. Now I get:
```
$ netcat localhost 3030 -
{"id":1,"jsonrpc":"2.0","method":"hello_subscribe","params":[10]}
{"jsonrpc":"2.0","result":0,"id":1}
{"jsonrpc":"2.0","method":"hello","params":{"result":"Hello World!","subscription":0}}
{"jsonrpc":"2.0","method":"hello","params":{"result":"Hello World!","subscription":0}}
[...]

PR #121, opened 2017-04-20
Fixing borked alias in macros

tomusdrw:
Related #120


rphmeier:
ICE on nightly doesn't seem good...

tomusdrw:
Yeah, let's wait and see if it's fixed by https://github.com/rust-lang/rust/issues/41298 
Otherwise we'll report another issue.

PR #122, opened 2017-04-20
Update/refactor to latest tokio_io

NikVolf:
Also added some tests which had been failing before refactoring, but now ok, since tokio guys resolved issues with handles recycling.

### DIFF
--- a/ipc/Cargo.toml
+++ b/ipc/Cargo.toml
@@ -14,6 +14,7 @@ tokio-service = "0.1"
 jsonrpc-core = { version = "7.0", path = "../core" }
 jsonrpc-server-utils = { version = "7.0", path = "../server-utils" }
 parity-tokio-ipc = { git = "https://github.com/nikvolf/parity-tokio-ipc" }
+bytes = "0.4"
 
 [dev-dependencies]
 env_logger = "0.4"
--- a/ipc/Cargo.toml
+++ b/ipc/Cargo.toml
@@ -16,7 +16,6 @@ jsonrpc-server-utils = { version = "7.0", path = "../server-utils" }
 parity-tokio-ipc = { git = "https://github.com/nikvolf/parity-tokio-ipc" }
 bytes = "0.4"
 
-
 [dev-dependencies]
 env_logger = "0.4"
 lazy_static = "0.2"
--- a/ipc/Cargo.toml
+++ b/ipc/Cargo.toml
@@ -16,6 +16,7 @@ jsonrpc-server-utils = { version = "7.0", path = "../server-utils" }
 parity-tokio-ipc = { git = "https://github.com/nikvolf/parity-tokio-ipc" }
 bytes = "0.4"
 
+
 [dev-dependencies]
 env_logger = "0.4"
 lazy_static = "0.2"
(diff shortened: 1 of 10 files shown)


NikVolf:
nightly compiler bug seems unrelated

ipc/src/logger.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(dead_code)]
```
tomusdrw: Why is it needed? Seems a bit too much to allow dead code in the whole module.

ipc/src/server.rs
```diff
@@ -109,7 +109,7 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
 
 				let meta = meta_extractor.extract(&RequestContext { endpoint_addr: &remote_id });
 				let service = Service::new(rpc_handler.clone(), meta);
-				let (writer, reader) = io_stream.framed(StreamCodec).split();
+				let (writer, reader) = ::server_utils::tokio_io::AsyncRead::framed(io_stream, StreamCodec).split();
```
tomusdrw: Why can't we just import the trait and use the code as previously?

ipc/src/server.rs
```diff
@@ -160,7 +160,7 @@ pub struct Server {
 impl Server {
 	/// Closes the server (waits for finish)
 	pub fn close(mut self) {
-		self.stop.take().unwrap().complete(());
+		self.stop.take().map(|stop| stop.send(()).unwrap_or_else(|_| { /* might be closed already */ } ));
```
tomusdrw: Why not `let _ = self.stop.take().map(|stop| stop.send(()));` or `let _ = self.stop.take().and_then(|stop| stop.send(()).ok());`?

tcp/src/server.rs
```diff
@@ -164,19 +163,19 @@ pub struct Server {
...
 	pub fn close(mut self) {
-		self.stop.take().unwrap().complete(());
+		self.stop.take().map(|sg| sg.send(()).unwrap_or_else(|_| {} /* might be already dropped */));
 		self.remote.take().unwrap().close();
 	}
 
 	/// Wait for the server to finish
 	pub fn wait(mut self) {
-		self.remote.take().unwrap().wait();
+		self.remote.take().map(|handle| handle.wait());
```
tomusdrw: Why is it possible that handle is not there? Seems that `close` always expects it?
(Would be good to have `.expect` with a proof here and couple of lines above instead of `unwrap`)

ipc/Cargo.toml
```diff
@@ -14,6 +14,8 @@ tokio-service = "0.1"
 jsonrpc-core = { version = "7.0", path = "../core" }
 jsonrpc-server-utils = { version = "7.0", path = "../server-utils" }
 parity-tokio-ipc = { git = "https://github.com/nikvolf/parity-tokio-ipc" }
+bytes = "0.4"
+
```
tomusdrw: Superfluous line.

ipc/src/logger.rs
NikVolf: it's just logger can be used or not in tests
generally it is just turned on when trying to figure out why tests are broken

NikVolf:
@tomusdrw 
resolved with warnings and others

PR #126, opened 2017-04-24
WS pubsub example.

tomusdrw:
- Exposing `futures::sync::mpsc::Sender` in `RequestContext`
- Apparently `ws::Sender` doesn't track if the connection is still active so I've added manual tracking and exposed it in `RequestContext`.
- Updates all examples to avoid glob imports (only `jsonrpc_core` is imported with a wildcard)

Closes: #120

### DIFF
--- a/http/examples/http_async.rs
+++ b/http/examples/http_async.rs
@@ -4,7 +4,7 @@ extern crate jsonrpc_http_server;
 
 use jsonrpc_core::*;
 use jsonrpc_core::futures::Future;
-use jsonrpc_http_server::*;
+use jsonrpc_http_server::{ServerBuilder, DomainsValidation, AccessControlAllowOrigin};
 
 fn main() {
 	let mut io = IoHandler::default();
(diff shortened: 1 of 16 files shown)


faern:
@tomusdrw Not using wildcard imports (`use jsonrpc_core::*;`) would make it easier for newcomers to track where types used in the example come from.

tomusdrw:
@faern Thanks for the suggestion. Updated all examples.

faern:
@tomusdrw Is the `add_method`/`add_subscription` or the macros approach the preferred usage of this crate? Are both meant to be supported for a long time? I got the impression that the macros support is the new kid on the block and was a bit surprised that the new example then uses the other way.

tomusdrw:
`pubsub` crate is more low-level (works on raw `Param`, `Value` and method level), since it's a custom extension to JSON-RPC protocol we want to keep it separate.

`macros` crate is designed to create a complete services on top of those low-level crates, but uses a bit of `macro` magic currently.

That said both methods are equally supported, our main driver for this library is development of RPC interfaces in https://github.com/paritytech/parity and we actually use `macros` there.

The example uses `add_method` stuff cause it's in `pubsub` crate, feel free to add another example in `macros` crate to show WS + PubSub + Macros.

faern:
Using macros instead was pretty easy. I don't think there is any information there that is not already available in the existing PubSub + Macros examples. Getting this merged as is with the added `Sender`/`RequestContext` functionality will satisfy my needs for now.

PR #129, opened 2017-04-28
Support RPC methods with 6 arguments

svyatonik:
To address this issue:
https://github.com/paritytech/parity/pull/5439#issuecomment-297962170


tomusdrw:
@svyatonik could you prepare a backport PR to `parity-1.7` branch?

PR #131, opened 2017-05-02
Fix IPC issue with responses > 8kb

NikVolf:
also contains some tests refactoring to allow them to use framed request-replies

### DIFF
--- a/ipc/src/server.rs
+++ b/ipc/src/server.rs
@@ -269,7 +269,7 @@ mod tests {
 		assert_eq!(
 			result,
 			"{\"jsonrpc\":\"2.0\",\"result\":\"hello\",\"id\":1}",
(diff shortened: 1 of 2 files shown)


ipc/src/server.rs
```diff
@@ -323,4 +320,48 @@ mod tests {
...
+		thread::sleep(::std::time::Duration::from_millis(50));
+
+		let result = dummy_request_str(&path,
+			"{\"jsonrpc\": \"2.0\", \"method\": \"say_huge_hello\", \"params\": [], \"id\": 1}",
+		);
+
+		assert_eq!(
+			result,
+			huge_response_test_json(),
+			"Response does not exactly much the expected response",
```
tomusdrw: `s/much/match`

PR #136, opened 2017-05-05
Supporting Host/Origin wildcards.

tomusdrw:


server-utils/src/matcher.rs
```diff
@@ -0,0 +1,48 @@
...
+		)
+	}
+
+	pub fn matches<T: AsRef<str>>(&self, other: T) -> bool {
+		let s = other.as_ref();
+		let res = match self.0 {
+			Some(ref matcher) => matcher.is_match(s),
+			None => self.1.eq_ignore_ascii_case(s),
+		};
+		println!("Checking: {} vs {} = {:?}", self.1, s, res);
```
maciejhirsz: Intended or left over?

tomusdrw: Fixed!

PR #142, opened 2017-05-11
Updating to mio-0.6

tomusdrw:


NikVolf:
Test fails seems legit, why do we need update mio in stable branch though?

tomusdrw:
Initially I thought it's required to get rid of `nix-0.5.0` but it turned out that only `0.5.0->0.5.1` update was needed. As I have almost finished this migration decided to make a PR.

tomusdrw:
Not needed any more.

PR #146, opened 2017-05-18
Query ws for actual address bound to when listening

faern:
This PR uses a newer version of `ws` in my fork of that repo. I suggest first updating https://github.com/tomusdrw/ws-rs to the code at https://github.com/faern/ws-rs/tree/arbitrary-handshake-responses. Then this PR does not have to change the dependency any longer and I can remove that commit. So that this repo never have to depend on my fork.

Other than than this PR uses the new `WebSocket::local_addr` method to query which `SocketAddr` we were actually bound to. Thus making it possible to find out which port was chosen if one bind to port `0` to get a random one.

### DIFF
--- a/ws/src/server.rs
+++ b/ws/src/server.rs
@@ -63,6 +63,7 @@ impl Server {
 		// Start listening...
 		let ws = ws.bind(addr)?;
 		let local_addr = ws.local_addr()?;
+		debug!("Bound to local address: {}", local_addr);
 
 		// Spawn a thread with event loop
 		let handle = thread::spawn(move || {
--- a/ws/src/server.rs
+++ b/ws/src/server.rs
@@ -62,6 +62,8 @@ impl Server {
 
 		// Start listening...
 		let ws = ws.bind(addr)?;
+		let local_addr = ws.local_addr()?;
+
 		// Spawn a thread with event loop
 		let handle = thread::spawn(move || {
 			match ws.run().map_err(Error::from) {
@@ -75,7 +77,7 @@ impl Server {
 
 		// Return a handle
 		Ok(Server {
-			addr: addr.to_owned(),
+			addr: local_addr,
 			handle: Some(handle),
 			remote: Some(eloop),
 			broadcaster: broadcaster,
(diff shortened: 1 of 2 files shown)


tomusdrw:
I think it's fine to keep it on your fork for now, I want to switch to crates.io version when the patched `ws` is released.

faern:
@tomusdrw Cool. Then go ahead and merge if you are happy with pointing to my fork.

About the remaining stuff before we can switch to the crates.io version of `ws`. It's the custom body in responses. I have tried to make that work for the client part of `ws`, but so far failed and thus I can't really create an automated test for it either. I doubt they want to merge that into `ws` if it's only one sided. This is not a very high priority for me at the moment, so I won't put any time into fixing this now.

tomusdrw:
@rphmeier Could you have a look as well?

ws/Cargo.toml
```diff
@@ -12,4 +12,4 @@ documentation = "https://paritytech.github.io/jsonrpc/json_ws_server/index.html"
 log = "0.3"
 jsonrpc-core = { version = "7.0", path = "../core" }
 jsonrpc-server-utils = { version = "7.0", path = "../server-utils" }
-ws = { git = "https://github.com/tomusdrw/ws-rs" }
+ws = { git = "https://github.com/faern/ws-rs", branch = "arbitrary-handshake-responses" }
```
rphmeier: since we use this crate as a git dependency in Parity, I think this would let you introduce arbitrary code changes into Parity master at will.

faern: Yup. This is one of the reasons I suggested you do that :) So you don't have to trust my repo.

tomusdrw: We still use a lock file in Parity, also this change would need to be ported to `parity-1.*` branch first. I think the worst that could happen is that the builds would be broken - but that's true regarding any dependency referenced from git.

Anyway the changes are merged to `tomusdrw/ws-rs`, @faern would you mind updating the PR to reference back this repo?

faern: Done. Removed that commit.

PR #163, opened 2017-07-03
Call sync methods lazily.

tomusdrw:


tomusdrw:
Reviewed by @rphmeier offline.

PR #165, opened 2017-07-03
Process all calls in a lazy way.

tomusdrw:


macros/src/delegates.rs
```diff
@@ -12,7 +12,7 @@ type AsyncData = BoxFuture<Value, Error>;
 
 struct DelegateMethod<T, F> {
 	delegate: Arc<T>,
-	closure: F,
+	closure: Arc<F>,
```
rphmeier: why not hold `T` and `F` by value but implement `RpcMethod` for `Arc<DelegateMethod<T, F>>`? Then only one clone has to be done per call

tomusdrw: Unfortunatelly `delegate` is shared between many methods, so `Arc<T>` is required anyway.

PR #166, opened 2017-07-03
Fix a panic caused by buf not being able to accommodate msg 

SammyJames:


tcp/src/line_codec.rs
```diff
@@ -28,7 +28,7 @@ impl Encoder for LineCodec {
 	type Error = io::Error;
 
 	fn encode(&mut self, msg: String, buf: &mut BytesMut) -> io::Result<()> {
-		buf.put_slice(msg.as_bytes());
+		buf.extend_from_slice(msg.as_bytes());
 		buf.put(b'\n');
```
NikVolf: just realized that on he edge case when msg.len() will exactly match the buffer size the code will panic here (since msg will occupy the whole buffer and there will be no place for `\n`)

care to make another pr to fix it or i will take it over?

PR #169, opened 2017-07-04
Merge ipc/tcp codecs to one, configurable

NikVolf:


server-utils/src/lib.rs
```diff
@@ -15,5 +17,12 @@ pub mod hosts;
 pub mod session;
 pub mod reactor;
 mod matcher;
+mod stream_codec;
 
 pub use matcher::Pattern;
+
+/// Codecs utilities
+pub mod codecs {
```
tomusdrw: Why not just make `stream_codec` module public?

server-utils/src/stream_codec.rs
```diff
@@ -1,8 +1,56 @@
-use std::io;
-use server_utils::tokio_io::codec::{Decoder, Encoder};
+use std::{io, str};
+use tokio_io::codec::{Decoder, Encoder};
 use bytes::BytesMut;
 
-pub struct StreamCodec;
+/// Separator for enveloping messages in streaming codecs
+pub struct Separator(Option<u8>);
```
tomusdrw: I think `Option`-like enum would do here, why do you need a separate struct that wraps it?

server-utils/src/lib.rs
NikVolf: Just wanted all codecs (currently only one) it accessible via short namespace `codecs`

server-utils/src/stream_codec.rs
NikVolf: so i could make a separate `impl Default` for it
along with other custom methods

NikVolf: oh you probably ment smth like 

```
enum Separator { Empty, Byte(u8) }
```
?

that makes sense

tomusdrw: Yeah.

server-utils/src/lib.rs
tomusdrw: k

server-utils/src/stream_codec.rs
NikVolf: Updated!

PR #175, opened 2017-07-23
Stop dispatching websocket messages once connection is known closed

NikVolf:


ws/src/session.rs
```diff
@@ -233,6 +233,10 @@ impl<M: core::Metadata, S: core::Middleware<M>> ws::Handler for Session<M, S> {
 	}
 
 	fn on_message(&mut self, msg: ws::Message) -> ws::Result<()> {
+		if !self.active.load(atomic::Ordering::SeqCst) {
+			return Err(ws::Error::new(ws::ErrorKind::Internal, "Attempting to send a message to closed/closing connection."))
+		}
```
rphmeier: not actually sending anything here, only receiving

NikVolf: yeah, moved to a handling future

ws/src/server_builder.rs
```diff
@@ -54,10 +57,11 @@ impl ::std::error::Error for Error {
...
 
     fn cause(&self) -> Option<&::std::error::Error> {
-        Some(match *self {
-            Error::Io(ref e) => e,
-            Error::WebSocket(ref e) => e,
-        })
+        match *self {
+            Error::Io(ref e) => Some(e),
+            Error::WebSocket(ref e) => Some(e),
+			Error::ConnectionClosed => None,
```
ngotchac: There might be a space/tab mismatch here

rphmeier: yea, it uses tabs like the rest of the codebase while the file (mistakenly) uses spaces

NikVolf: see https://github.com/paritytech/jsonrpc/pull/177

PR #178, opened 2017-07-26
keep_alive support

ekicyou:
add json-http-server::ServerBuilder::keep_alive implement.

from hyper::server::Http::keep_alive (https://docs.rs/hyper/0.11.1/hyper/server/struct.Http.html#method.keep_alive)

### DIFF
--- a/http/src/lib.rs
+++ b/http/src/lib.rs
@@ -310,8 +310,8 @@ impl<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>> ServerBuilder<M, S> {
 			handler: self.handler,
 			extractor: self.meta_extractor,
 		};
-		let reuse_port = self.threads > 1;
 		let keep_alive = self.keep_alive;
+		let reuse_port = self.threads > 1;
 
 		let (local_addr_tx, local_addr_rx) = mpsc::channel();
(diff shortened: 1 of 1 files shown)


http/src/lib.rs
```diff
@@ -407,7 +420,11 @@ fn serve<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>>(
 		bind_result.and_then(move |(listener, local_addr)| {
 			let allowed_hosts = server_utils::hosts::update(allowed_hosts, &local_addr);
 
-			let http = server::Http::new();
+			let http = {
```
NikVolf: can be just `let http = server::Http::new().keep_alive(keep_alive);` ?

ekicyou: [hyper::server::Http::keep_alive](https://docs.rs/hyper/0.11.1/hyper/server/struct.Http.html#method.keep_alive), result is ```&mut Self``` not ```Self```.

```rust
fn keep_alive(&mut self, val: bool) -> &mut Self
```

Because I want ```Self```, I create ```let mut Self```.

NikVolf: sorry, didn't check the docs
strange variant for using in api from hyper guys though :)

PR #195, opened 2017-08-29
Implement Debug

faern:
I tried to use `unwrap_err` on a local object containing a `jsonrpc-ws-server::Server`, but failed because it did not implement Debug. So I thought it would be helpful to add a `Debug` impl for that type and many other types I found in these crates.

Also fixed a few compiler warnings completely unrelated to `Debug`.


debris:
Neat!

PR #196, opened 2017-09-01
Adapt to Futures 0.1.15 by not using boxed()

faern:
Futures 0.1.15 deprecates the `BoxFuture` type and the `.boxed()` method on `Future`. The [issue](https://github.com/alexcrichton/futures-rs/issues/228) talking about that recommend that people use `Box::new(...)` instead because it allows the compiler to figure out some magic better. So I went ahead and fixed that.

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -49,7 +49,7 @@ use jsonrpc_core::futures::Future;
 fn main() {
 	let io = IoHandler::new();
 	io.add_async_method("say_hello", |_params: Params| {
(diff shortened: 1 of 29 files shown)


faern:
The test that fails on Travis works on my machine on stable Rust. Maybe just restart the build?

faern:
Ultimately it would probably be better if there could be less trait objects and more monomorphization. But I didn't feel this was the time to look at that as well :)

PR #197, opened 2017-09-03
http: handle charset=utf-8 for application/json

sorpaas:
Sometimes the browser would send content type as `application/json; charset=utf-8`. `jsonrpc` sees it is not equal to `application/json`, and reject that request.

In the mime crate there are also constants like `TEXT_PLAIN_UTF_8`, so I think using `APPLICATION_JSON_UTF_8` is the right way to handle this.


tomusdrw:
@sorpaas Thanks for the PR, would you mind writing a test in `http/src/tests.rs` for that particular case?

PR #199, opened 2017-09-04
Remove boxing and allow IntoFuture

### LINKED ISSUE #24: Macros: trait definition should support async or sync method impl
We can accommodate old uses of `Result` with `try!` using a [Carrier](https://doc.rust-lang.org/stable/std/ops/trait.Carrier.html) implementation and `?`. 

Until `Carrier` is stabilized we could just hack around it when porting existing synchronous method implementations
```rust
fn previous_sync(&self) -> BoxFuture<Success, Error> {
    let body = move || {
        // previous function body here
    };

    Future::done(body()).boxed()
}
```

tomusdrw:
- Changes the interface to allow RPC methods to return `IntoFuture` in many places.
- Reduces boxing (now happens almost exclusively within `RpcMethod`)
- Removes sync/async distinction (Resolves #24)


core/src/calls.rs
```diff
@@ -1,22 +1,18 @@
...
-/// Synchronous Method
-pub trait RpcMethodSync: Send + Sync + 'static {
-	/// Call method
-	fn call(&self, params: Params) -> Result<Value, Error>;
-}
-
 /// Asynchronous Method
 pub trait RpcMethodSimple: Send + Sync + 'static {
+	/// Output future
+	type Out: Future<Item = Value, Error = Error> + Send;
```
rphmeier: why not bound this by `IntoFuture` also?

rphmeier:
also seems to remove the distinction between sync and async methods as a side effect? either way, super cool! one fewer allocation per RPC call and it opens the door to having `impl Future` RPC methods

core/src/calls.rs
tomusdrw: 1. Consistency: `RpcMethod:call` returns `Future` as well
2. `Send` bound

http/src/handler.rs
```diff
@@ -46,37 +46,44 @@ impl<M: Metadata, S: Middleware<M>> server::Service for ServerHandler<M, S> {
 	type Future = Handler<M, S>;
 
 	fn call(&self, request: Self::Request) -> Self::Future {
-		let action = self.middleware.on_request(&request);
+		enum MiddlewareResponse {
```
rphmeier: usually, `Result` is used for cases like this. I recommend a typedef

PR #201, opened 2017-09-06
Correct stream codec handling of escape codes. Closes #200.

### LINKED ISSUE #200: jsonrpc-ipc-server hangs when sending strings containing quotes
If a request sent to a jsonrpc-ipc-server contains a double quote (`"`), correctly escaped in JSON, the server does not respond to the request.

A minimal example of the issue is as follows:

```rust
use jsonrpc_core::types::Value;
use jsonrpc_core::IoHandler;
use jsonrpc_ipc_server::ServerBuilder;
use std::io::{BufRead, BufReader, Write};
use std::os::unix::net::UnixStream;

let mut ioh = IoHandler::new();
ioh.add_method("foo", |_| Ok(Value::Null));
let srv = ServerBuilder::new(ioh).start("/tmp/foo.sock").unwrap();
let mut stream = UnixStream::connect("/tmp/foo.sock").unwrap();
let mut line = String::new();
// Without double quote: OK
stream.write_all(r#"{"jsonrpc":"2.0","method":"foo","id":0,"params":["foobar"]}"#.as_bytes()).unwrap();
BufReader::new(&mut stream).read_line(&mut line).unwrap();
print!("{}", line);
// With double quote: Hangs
stream.write_all(r#"{"jsonrpc":"2.0","method":"foo","id":1,"params":["foo\"bar"]}"#.as_bytes()).unwrap();
BufReader::new(&mut stream).read_line(&mut line).unwrap();
print!("{}", line);
srv.close();
```

The server will respond to the first request, but not the second. I have not tested if this issue affects other server implementation, however it is not an issue in serde or jsonrpc-core.

tkerber:


server-utils/src/stream_codec.rs
```diff
@@ -87,12 +87,14 @@ impl Decoder for StreamCodec {
 				else if byte == b'"' && !is_escaped {
 					in_str = !in_str;
 				}
-				else if byte == b'\\' && is_escaped && !in_str {
-					is_escaped = !is_escaped;
-				}
 				else if is_whitespace(byte) {
 					whitespaces += 1;
 				}
+				if byte == b'\\' && !is_escaped && in_str {
```
tomusdrw: should be started with `else` as well

tkerber: I didn't use `else if` here, as I wanted `is_escaped` to be unset in all cases. With `else if`, it wouldn't be unset if a whitespace followed the backslash. That said, it wouldn't be correctly formed JSON if any of the other cases are caught.

tomusdrw: Oh, you're absolutely right. Could you add more test cases covering these cases?

NikVolf:
Thanks!

PR #202, opened 2017-09-07
Impl from on error kind

faern:
There was no way to instantiate an `ErrorKind` without deserializing it. So I moved that code to a `From<i64>` impl and made use of it in the deserialization.

I also realized that deserializing an `ErrorCode` from something that was a valid json `Value`, but not an `i64` would trigger the `unimplemented!()` call. Panics are dangerous, especially in exposed APIs. I changed that and hopefully this should never panic.


debris:
This bug was really stupid. Thanks for the fix! @tomusdrw I will release new version of the library with this fix and I will backport it to the parity-1.7 branch

tomusdrw:
@debris no point in backporting to 1.7, I'm migrating Parity to use latest master currently.

debris:
great! :)

PR #206, opened 2017-10-06
Re-export hyper::server::Request

faern:
Re-export `hyper::server::Request` in `jsonrpc_http_server` so that any user of this library who wants to write a middleware or meta extractor does not have to explicitly depend on and import hyper by themselves. Will make it easier to use.

### DIFF
--- a/http/src/lib.rs
+++ b/http/src/lib.rs
@@ -49,6 +49,7 @@ pub use server_utils::tokio_core;
 pub use handler::ServerHandler;
 pub use utils::{is_host_allowed, cors_header, CorsHeader};
 pub use response::Response;
+pub use hyper::server::Request;
 
 /// Result of starting the Server.
 pub type ServerResult = Result<Server, Error>;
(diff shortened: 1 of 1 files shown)


tomusdrw:
Shouldn't we re-export `Response` as well for completeness? Or maybe it would be better to re-export whole hyper?

faern:
I didn't notice `hyper::server::Response` was used publicly. But it is. So yes, it should be re-exported. But it collides with the local `Response`, so one of them must change name.

I think re-exporting all of hyper is overkill, one does not need the rest in order to interact with this library.

faern:
Could be `pub use hyper::server::{Request as HyperRequest, Reply as HyperReply};` or do you prefer just re-exporting all of hyper?

tomusdrw:
I'd go with re-exporting all of hyper, cause if you for instance want to check a header on request you will need it anyway, same with creating a response.

faern:
Oh wait.. Hyper already is publicly exported here. My bad! This PR is void :)

PR #208, opened 2017-10-16
REST -> RPC converter for HTTP server

tomusdrw:
Allows to send simple RPC requests using for instance `curl`:
```
curl -X POST http://localhost:3000/<method>/<param1>
```
Which get's converted to an actual request:
```
{ "id": 1, "jsonrpc": "2.0", "method": <method>, "params": [<param1>]}
```


tomusdrw:
@debris did you manage to review it?

PR #211, opened 2017-10-29
Add helper methods to Output

faern:
Handy methods to get the `Id` and `Version` from an `Output` without having to pattern match every time. Very useful for checking incoming `Output` objects that they have the correct id and version before handling the actual content.

Also added the `into_result` since it felt useful.

### DIFF
--- a/core/src/types/response.rs
+++ b/core/src/types/response.rs
@@ -78,10 +78,12 @@ impl Output {
 			Output::Failure(ref f) => &f.id,
 		}
 	}
+}
 
+impl From<Output> for Result<Value, Error> {
 	/// Convert into a result. Will be `Ok` if it is a `Success` and `Err` if `Failure`.
-	pub fn into_result(self) -> Result<Value, Error> {
-		match self {
+	fn from(output: Output) -> Result<Value, Error> {
+		match output {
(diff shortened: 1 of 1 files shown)


core/src/types/response.rs
```diff
@@ -62,6 +62,30 @@ impl Output {
...
+	/// Get the correlation id.
+	pub fn id(&self) -> &Id {
+		match *self {
+			Output::Success(ref s) => &s.id,
+			Output::Failure(ref f) => &f.id,
+		}
+	}
+
+	/// Convert into a result. Will be `Ok` if it is a `Success` and `Err` if `Failure`.
+	pub fn into_result(self) -> Result<Value, Error> {
```
tomusdrw: Why not `Into<Result<V, E>>`? or better `From<Output> for Result<V, E>`?

faern: Ah, of course. Fixing instantly.

tomusdrw:
@debris can you have a look as well?

faern:
Was that build error triggered by my code? Seems unlikely, maybe restart that build?

debris:
lgtm, thanks! ;)

PR #215, opened 2017-11-12
Introduce error_chain

### LINKED ISSUE #183: Use error-chain for error handling

tomusdrw:
Closes #183


http/src/error.rs
```diff
@@ -0,0 +1,28 @@
...
+use hyper;
+
+error_chain! {
+	foreign_links {
+		Io(io::Error);
+	}
+
+	errors {
+		/// Hyper error
+		Hyper(e: hyper::Error) {
```
debris: why not `foreign_links`? It [implements](https://docs.rs/hyper/0.11.6/hyper/error/enum.Error.html#impl-StdError) `std::error::Error`

ws/src/error.rs
```diff
@@ -0,0 +1,36 @@
...
+
+	errors {
+		/// Attempted action on closed connection.
+		ConnectionClosed {
+			description("connection is closed"),
+			display("Action on closed connection."),
+		}
+
+		/// WebSockets error.
+		WebSocket(t: ws::Error) {
```
debris: why not `foreign_links`?

http/src/error.rs
tomusdrw: Explanation given in custom `From` implementation. We want to unwrap `Io` error from hyper error.

ws/src/error.rs
tomusdrw: Same as above.

http/src/error.rs
debris: but why?

Issue #179, opened 2017-08-04
What is metadata?


jtremback:
I don't see any documentation here or in the jsonrpc spec on "metadata", and what its role is.

tomusdrw:
`Metadata` is a custom extension of JSONRPC spec. It allows a method handler to access some additional data from the transport layer, which is not part of the payload itself.

For instance you can check a HTTP header and alter the behaviour (obviously it's against transport-agnostic JSONRPC spec). More real-life example would be to check authorization header presence in `jsonrpc-core` middleware.

It's also used by our PubSub mechanism (also a custom extension), where metadata contains a pipe allowing the method handlers to send notifications to the clients.

`jsonrpc_core::IoHandler` is a pure implementation of JSONRPC spec. This extension is available in `jsonrpc_core::MetaIoHandler`.

Let me know if you need any more detailed info on this.

greenpdx:
I need some help with Metadata, If I could get just a simple template I could do it. This is what I have now

''' rust
extern crate jsonrpc_core;
extern crate jsonrpc_http_server;

use jsonrpc_core::*;
use jsonrpc_http_server::{ServerBuilder, DomainsValidation, AccessControlAllowOrigin, RestApi, MetaExtractor};

#[derive(Clone, Default)]
struct Meta(usize);
impl MetaExtractor<Meta> for Meta {
    fn read_metadata(&self, req: &server::Request) -> Meta {
        (*self)(req)
    }
}

fn main() {
    let mut io = MetaIoHandler::default();
    io.add_method_with_meta("say_hello", |_parm: Params, meta: Meta| {
        println!("Hello {:?}", _parm );
        futures::finished(Value::String(format!("hello: {}", meta.0)))
        //Ok(Value::String("hello".to_string()))
    });

    let _server = ServerBuilder::new(io)
        .threads(3)
        .rest_api(RestApi::Unsecure)
        .meta_extractor(Meta::read_metadata())
        .cors(DomainsValidation::AllowOnly(vec![AccessControlAllowOrigin::Any]))
        .start_http(&"0.0.0.0:3030".parse().unwrap())
        .expect("unable to start");

    _server.wait();
}

'''

tomusdrw:
@greenpdx Please have a look at the example I added here: https://github.com/paritytech/jsonrpc/pull/219

greenpdx:
I am testing it using Vue, send it a json request with params
[...]

PR #219, opened 2017-11-24
Add http example with metadata.

tomusdrw:
Related #179


tomusdrw:
@greenpdx Just use other header from `hyper`, for instance: https://docs.rs/hyper/0.11.7/hyper/header/struct.Host.html. You can also use `Headers::get_raw` to retrieve any header you want in it's raw form.

greenpdx:
Great, thanks, I am now working on getting rpc commands for opencv 3.3.

PR #221, opened 2017-11-27
Get rid of Default bound for Metadata.

tomusdrw:
Whole `Metadata` doesn't really require `Default` bound, it's only needed to use some methods from `IoHandler`, so the bound has been moved there.

### DIFF
--- a/core/examples/middlewares.rs
+++ b/core/examples/middlewares.rs
@@ -5,7 +5,7 @@ use std::sync::atomic::{self, AtomicUsize};
 use jsonrpc_core::*;
 use jsonrpc_core::futures::Future;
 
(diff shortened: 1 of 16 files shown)


ivanovaleksey:
Hello, I was struggling with this :)
I got an error 
```
the trait `std::default::Default` is not implemented for `Meta`
```
but I couldn't understand why I was receiving it.

Suddenly I realized that I was looking at code on _master_ at my application was using version from _crates.io_.
Maybe we could push a new release?

ivanovaleksey:
@tomusdrw just mention you in case you don't receive a notification.

tomusdrw:
@ivanovaleksey thanks. We did discuss that with @debris and going to release soon.

PR #222, opened 2017-12-13
Fix Params deserialization on serde_json>=1.0.8

Marwes:
Due to https://github.com/serde-rs/json/pull/389, serde_json no longer calls `deserialize_any` in `deserialize_identifer` (and all other default implemented methods). This causes `Params` to fail to deserialize maps and sequences since it has hinted that it expects an identifier.

This implements `deserialize` correctly, saying that it accepts any type.

cc @dtolnay (thought you might want to be aware of this regression/bug)

### DIFF
--- a/core/src/types/id.rs
+++ b/core/src/types/id.rs
@@ -29,7 +29,7 @@ impl Serialize for Id {
 impl<'a> Deserialize<'a> for Id {
 	fn deserialize<D>(deserializer: D) -> Result<Id, D::Error>
 	where D: Deserializer<'a> {
-		deserializer.deserialize_identifier(IdVisitor)
+		deserializer.deserialize_any(IdVisitor)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


tomusdrw:
Hi @Marwes! Thanks a bunch. Would you mind adding a test to cover current incorrect behaviour?

Marwes:
It should be covered by https://github.com/Marwes/jsonrpc/blob/9d6fcf1aa9d4f3b2eb174f83b4785284501976f8/core/src/types/params.rs#L98-L110 . It is just that the last build of jsonrpc-core was with serde_json<1.0,8.

(Also I couldn't build this locally due to https://github.com/rust-lang/cargo/issues/4810 so I just opted for the minmal change through the github gui :/ )

Marwes:
Looks to be building with 1.0.8 so if it passes this should be good https://travis-ci.org/paritytech/jsonrpc/jobs/316075816#L491

tomusdrw:
I see, thanks! Will try to bump in my local Cargo.lock and test.

Marwes:
Same issue for `Id`. Fixed.

debris:
@tomusdrw we need to introduce the same fixes in [parity](https://github.com/paritytech/parity)

PR #227, opened 2018-01-18
removed warnings related to unused AsciiExt

debris:


server-utils/src/matcher.rs
```diff
@@ -1,5 +1,4 @@
 use globset::{GlobMatcher, GlobBuilder};
-use std::ascii::AsciiExt;
```
mmacedoeu: matcher still using trait:

error[E0599]: no method named `eq_ignore_ascii_case` found for type `std::string::String` in the current scope
  --> /home/macedo/.cargo/git/checkouts/jsonrpc-22a65938f7d26b63/d1e2cc4/server-utils/src/matcher.rs:31:19
   |
31 |    None => self.1.eq_ignore_ascii_case(s),
   |                   ^^^^^^^^^^^^^^^^^^^^
   |
   = help: items from traits can only be used if the trait is in scope
   = note: the following trait is implemented but not in scope, perhaps add a `use` for it:
           candidate #1: `use std::ascii::AsciiExt;`

tomusdrw: @mmacedoeu What's your `rustc --version`? Could you please check if it works correctly on latest stable?

tomusdrw: Please try on the latest stable:
```
$ rustc --version
rustc 1.23.0 (766bd11c8 2018-01-01)
```

mmacedoeu: Works with 1.23.0

PR #238, opened 2018-02-16
Add scope for eq_ignore_ascii_case

### LINKED ISSUE #237: Failing to build parity
Parity build using `cargo build --release` fails with following error:

```
error[E0599]: no method named `eq_ignore_ascii_case` found for type `std::string::String` in the current scope
  --> /Users/ayushyachitransh/.cargo/git/checkouts/jsonrpc-22a65938f7d26b63/786f3e1/server-utils/src/matcher.rs:31:19
   |
31 |    None => self.1.eq_ignore_ascii_case(s),
   |                   ^^^^^^^^^^^^^^^^^^^^
   |
   = help: items from traits can only be used if the trait is in scope
   = note: the following trait is implemented but not in scope, perhaps add a `use` for it:
           candidate #1: `use std::ascii::AsciiExt;`
```

Before building, it has been ensured that I have the latest code from the repository of paritytech/parity and the crate being used is the latest available on master of `paritytech/jsonrpc`

AyushyaChitransh:
Resolves #237

### DIFF
--- a/server-utils/src/matcher.rs
+++ b/server-utils/src/matcher.rs
@@ -1,4 +1,5 @@
 use globset::{GlobMatcher, GlobBuilder};
+use std::ascii::AsciiExt;
 use std::{fmt, hash};
 
 /// Pattern that can be matched to string.


tomusdrw:
The issue is resolved by updating compiler to latest stable version. On newer versions it generates a warning:
```
warning: unused import: `std::ascii::AsciiExt`
 --> server-utils/src/matcher.rs:2:5
  |
2 | use std::ascii::AsciiExt;
  |     ^^^^^^^^^^^^^^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default
```

PR #243, opened 2018-03-14
Disable fragment grow and limit in_buffer.

tomusdrw:


ws/src/server.rs
```diff
@@ -52,8 +52,14 @@ impl Server {
 	) -> Result<Server> {
 		let config = {
 			let mut config = ws::Settings::default();
+			// don't grow non-final fragments (to prevent DOS)
+			config.fragments_grow = false;
+			// don't accept super large requests
+			config.max_in_buffer = 5 * 1024 * 1024; // 5MB
```
folsen: Is this per buffer frame or the total for the whole request?

tomusdrw: It's the `in_buffer` for the entire connection.

PR #249, opened 2018-03-21
Only process request bodies up to a maximum size.

twittner:
Return 400 response if request body size exceeds maximum.

### DIFF
--- a/http/src/handler.rs
+++ b/http/src/handler.rs
@@ -1,6 +1,6 @@
 use Rpc;
 
-use std::{fmt, mem};
+use std::{fmt, mem, str};
 use std::sync::Arc;
 
 use hyper::{self, mime, server, Method};
@@ -22,6 +22,7 @@ pub struct ServerHandler<M: Metadata = (), S: Middleware<M> = NoopMiddleware> {
 	cors_domains: CorsDomains,
 	middleware: Arc<RequestMiddleware>,
 	rest_api: RestApi,
+	max_request_body_size: usize,
 }
 
 impl<M: Metadata, S: Middleware<M>> ServerHandler<M, S> {
@@ -32,6 +33,7 @@ impl<M: Metadata, S: Middleware<M>> ServerHandler<M, S> {
 		allowed_hosts: AllowedHosts,
 		middleware: Arc<RequestMiddleware>,
 		rest_api: RestApi,
+		max_request_body_size: usize,
 	) -> Self {
 		ServerHandler {
 			jsonrpc_handler,
@@ -39,6 +41,7 @@ impl<M: Metadata, S: Middleware<M>> ServerHandler<M, S> {
 			cors_domains,
 			middleware,
 			rest_api,
+			max_request_body_size,
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


http/src/handler.rs
```diff
@@ -319,11 +350,17 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
...
-	) -> Result<RpcPollState<M, S::Future>, hyper::Error> {
+	) -> Result<RpcPollState<M, S::Future>, BodyError> {
 		loop {
 			match body.poll()? {
-				// TODO [ToDr] reject too large requests?
 				Async::Ready(Some(chunk)) => {
+					// first check if this chunk itself is too large (in case it is already
+					// equal to `usize::MAX` where addition would overflow)
+					if chunk.len() > self.max_request_body_size
+						|| request.len() + chunk.len() > self.max_request_body_size
```
tomusdrw: It can still overflow if `max_request_body_size` is set to a large value?
Perhaps just use `checked_add` instead?

http/src/handler.rs
```diff
@@ -189,7 +194,19 @@ impl<M: Metadata, S: Middleware<M>> Future for RpcHandler<M, S> {
...
 			RpcHandlerState::ReadingBody { body, request, metadata, uri, } => {
-				self.process_body(body, request, uri, metadata)?
+				match self.process_body(body, request, uri, metadata) {
+					Err(BodyError::Utf8(ref e)) => {
+						let mesg = format!("utf-8 encoding error at byte {} in request body", e.valid_up_to());
+						let resp = Response::bad_request(mesg);
+						RpcPollState::Ready(RpcHandlerState::Writing(resp))
+					}
+					Err(BodyError::TooLarge) => {
+						let resp = Response::bad_request("request body size exceeds allowed maximum");
```
tomusdrw: Might be better to return `413 PAYLOAD TOO LARGE` instead.

http/src/handler.rs
```diff
@@ -189,7 +194,19 @@ impl<M: Metadata, S: Middleware<M>> Future for RpcHandler<M, S> {
 				RpcPollState::Ready(self.read_headers(request, continue_on_invalid_cors))
 			},
 			RpcHandlerState::ReadingBody { body, request, metadata, uri, } => {
-				self.process_body(body, request, uri, metadata)?
+				match self.process_body(body, request, uri, metadata) {
+					Err(BodyError::Utf8(ref e)) => {
+						let mesg = format!("utf-8 encoding error at byte {} in request body", e.valid_up_to());
```
tomusdrw: Perhaps `415 Unsupported Media Type`? But I'm fine with 400 as well.

> The 415 (Unsupported Media Type) status code indicates that the
origin server is refusing to service the request because the payload
is in a format not supported by this method on the target resource.
The format problem might be due to the request's indicated
Content-Type or Content-Encoding, or as a result of inspecting the
data directly.

http/src/handler.rs
twittner: True. I was assuming, no one would deviate too much from the 5 MiB default, but `checked_add` is surely the better approach.

http/src/handler.rs
twittner: Agreed.

http/src/handler.rs
twittner: Since we already verified the content-type is `application/json` or `application/json; charset=utf-8` and [RFC8259](https://tools.ietf.org/html/rfc8259) states:

>  JSON text exchanged between systems that are not part of a closed    ecosystem MUST be encoded using UTF-8

I think a 400 is appropriate.

PR #265, opened 2018-05-04
Set Access-Control-Max-Age header to 600 to prevent every request having an OPTIONS request.

ethernomad:
https://github.com/paritytech/parity/issues/8542


tomusdrw:
Parity is using `jsonrpc-http-server` (http dir) not `minihttp`. Would also be awesome to have that option configurable.

PR #266, opened 2018-05-04
Set Access-Control-Max-Age header to 600 to prevent every request having an OPTIONS request.

ethernomad:
How about this one?

https://github.com/paritytech/parity/issues/8542
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age
https://stackoverflow.com/questions/17432982/should-i-expect-options-preflight-with-all-cors-requests


ethernomad:
Can this be merged yet?

Issue #66, opened 2017-03-09
Use procedural macros


tomusdrw:
Replace `build_rpc_trait` macro with custom derive since it starts to get unmaintanable.

ivanovaleksey:
Hello @tomusdrw, I would like to figure out how `build_rpc_trait` works and I am interested in this task.
Is it still actual? I could try to help although for now I got very few knowledge.

tomusdrw:
Yes, it's not super urgent, but probably worth looking into. Let me know if I can help you with that (ping me on gitter).

PR #267, opened 2018-05-18
Associated types in RPC traits

gavofyork:


rphmeier:
The `ASSOCIATED` type is strange to me. For the use-cases where I need this, it would be better with regular generics.

rphmeier:
I think by swapping the rules for traits with metadata and those without, we can get rid of the `ASSOCIATED` keyword in this.

tomusdrw:
@gavofyork @rphmeier I believe it's not required any more in Polkadot, is it? Would be awesome to avoid `ASSOCIATED` keyword if possible, see https://github.com/paritytech/jsonrpc/pull/267#issuecomment-394734135

rphmeier:
@tomusdrw no, not needed for polkadot any more. 100% agree about ASSOCIATED keyword.

tomusdrw:
Closing, since I doesn't seem to be needed. We might consider this as part of #66

PR #273, opened 2018-06-02
initial commit to confirm location of issue

cooganb:
This is in regards to [issue #6616](https://github.com/paritytech/parity/issues/6616) filed in `paritytech/parity`, but actually traces back to this dependency. 

From the issue:

```
For maximum compatibility, Parity should echo back all the headers from Access-Control-Request-Headers in Access-Control-Allow-Headers. There could be a blacklist of headers that web pages should be prohibited from sending, but I don't think Parity trusts the RPC headers at all anyway, so I can't think of a reason not to just let pages send whatever they want.
```

@tomusdrw I wanted to check with you (and whoever else) to make sure I'm correct in the placement of the new code and the tests that need to be adjusted / passed to make sure my code will play well with others.

### DIFF
--- a/server-utils/src/cors.rs
+++ b/server-utils/src/cors.rs
@@ -102,6 +102,9 @@ impl ops::Deref for Origin {
 
 /// Origins allowed to access
 #[derive(Debug, Clone, PartialEq, Eq)]
+
+/// 1) Adding new CORS protocol to enum below
+
 pub enum AccessControlAllowOrigin {
 	/// Specific hostname
 	Value(Origin),
@@ -274,6 +277,8 @@ mod tests {
 		assert_eq!(res, CorsHeader::NotRequired);
 	}
 
+// 2) This test needs to pass when new CORS requirements added
+
 	#[test]
 	fn should_return_domain_when_all_are_allowed() {
 		// given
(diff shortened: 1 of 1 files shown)


server-utils/src/cors.rs
```diff
@@ -102,6 +102,9 @@ impl ops::Deref for Origin {
 
 /// Origins allowed to access
 #[derive(Debug, Clone, PartialEq, Eq)]
+
+/// 1) Adding new CORS protocol to enum below
```
tomusdrw: I think it should rather be a separate enum as outlined in https://github.com/paritytech/parity/issues/6616#issuecomment-337229845

If we want to be super strict the headers could be defined on the `Origin` level, but IMHO global setting for all allowed Origins are fine.

So the steps would be:
1. Add new enum (in that file most likely)
2. Wherever `get_cors_header` is used we should also compare the headers in the request with the allowed list. If that matches and we send CORS headers we should also add `Access-Control-Allowed-Headers` to the response.

server-utils/src/cors.rs
```diff
@@ -102,6 +102,7 @@ impl ops::Deref for Origin {
 
 /// Origins allowed to access
 #[derive(Debug, Clone, PartialEq, Eq)]
+
```
rphmeier: remove this line

server-utils/src/cors.rs
```diff
@@ -133,6 +134,14 @@ impl<T: Into<String>> From<T> for AccessControlAllowOrigin {
 
 /// CORS Header Result.
 #[derive(Debug, Clone, PartialEq, Eq)]
+
```
rphmeier: and this

tomusdrw:
Closing as stale.

PR #277, opened 2018-06-06
fixed ipc leak

### LINKED ISSUE #275: IPC leaks connections (when internally using PubSubSession)
Heyah,

investigating a [leak in IPC connections in Parity](https://github.com/paritytech/parity/issues/8774) I was able to pin-point the location to the way that IPC-Server and PubSubSession are interacting leading to an internal circle of never ending futures. Which in turn means we are keeping those connections around even after they've terminated, keep the socket open and - ultimatily - leak those connections as well as memory, causing [Parity to be oom-killed](https://github.com/paritytech/parity/issues/8618) (at least I suspect this error to be the cause of that error. Can't rule out there might be another leak).

[I've commited a short example triggering that behaviour, together with some code comments](https://github.com/gnunicorn/jsonrpc/commit/911ea0d450d36dfc61d8da70e61e9545a850e782) of the cause (IMHO) [in this branch](https://github.com/gnunicorn/jsonrpc/tree/leaking-ipc-connections-example).

To reproduce this bug, check out the [`leaking-ipc-connections-example` branch](https://github.com/gnunicorn/jsonrpc/tree/leaking-ipc-connections-example), cd into `ipc` and run

`RUST_LOG=ipc=trace cargo run --example with-pubsubsession`

Now have something connect to the open socket `/tmp/json-ipc-test.ipc` and immediately close again or die, like so:
```bash
#!/bin/sh -e
for i in `seq 0 1000`
do
    echo $i
    nc localhost 30303 </dev/null >/dev/null &
    pid=$!
    sleep 0.1
    kill $pid
done;
```

You'll see a bunch of messages appear in the example looking something like
```
    Finished dev [unoptimized + debuginfo] target(s) in 4.23 secs
     Running `/home/ben/dev/parity/jsonrpc/target/debug/examples/with-pubsubsession`
TRACE 2018-06-05T10:33:41Z: jsonrpc_ipc_server::server: Accepted incoming IPC connection: 1
TRACE 2018-06-05T10:33:41Z: with_pubsubsession: ext
TRACE 2018-06-05T10:33:41Z: jsonrpc_ipc_server::server: Accepted incoming IPC connection: 2
TRACE 2018-06-05T10:33:41Z: with_pubsubsession: ext
TRACE 2018-06-05T10:33:41Z: jsonrpc_ipc_server::server: Accepted incoming IPC connection: 3
TRACE 2018-06-05T10:33:41Z: with_pubsubsession: ext
```

Notice how the [closing message `Peer: service finished`](https://github.com/paritytech/jsonrpc/blob/7adb1ae60ffb5872976a924d0f95e93b60ef775e/ipc/src/server.rs#L181) is absent -- connections aren't dropped even though the peer is gone.

You can also watch it grow the open file handlers with lsof (e.g. ```while true; do echo "`date +%s` : `lsof -c with-pubs | wc -l`"; sleep 5; done```)

debris:
closes #275 
closes https://github.com/paritytech/parity/issues/8774
~https://github.com/paritytech/parity/issues/8618~

The leak existed, cause we were merging incoming message responses stream with outgoing messages. If either of them existed, we were not dropping a connection.

To fix the issue, we finish the merged stream as soon as one of the merged streams ends.


debris:
thanks @gnunicorn for debugging the issue!

ipc/src/select_both.rs
```diff
@@ -0,0 +1,73 @@
...
+
+		match a.poll()? {
+			Async::Ready(Some(item)) => return Ok(Some(item).into()),
+			Async::Ready(None) => return Ok(None.into()),
+			Async::NotReady => (),
+		};
+
+		self.flag = !self.flag;
+
+		match b.poll()? {
```
svyatonik: Am I right that this whole match can be replaced with simple `b.poll()` call? :)

debris: 🤦‍♂️ 👍

gnunicorn:
> closes paritytech/parity#8774
> closes paritytech/parity#8618

Just pointing out, that this doesn't come into effect until [the cargo.toml branch references are upgraded, too as they don't point to master](https://github.com/paritytech/parity/blob/bc2f86e806f70d23c1b5123c43be0d4cc73819bc/rpc/Cargo.toml#L31-L36) (I suspect this ther is a branching whenever there is a release?).

debris:
@gnunicorn I know, we need to backport these fixes to that branch

debris:
hold on with merging this pr, tests are randomly failing. Most likely because `server` is now released before the end of the test

ipc/src/server.rs
```diff
@@ -172,7 +173,7 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
 					})
 				})
 				.filter_map(|x| x)
-				.select(receiver.map_err(|e| {
+				.select_both(receiver.map_err(|e| {
```
dvdplm: I for one would appreciate a comment as to why we're using `select_both` here – doesn't seem super-obvious! ;)

debris:
Io it seems like tests are failing, because every request uses it's own, new event loop. And order of the event loop initialization is not guaranteed. 

So sometimes tests fail, because server hasn't started yet or started and was already dropped (because nothing has an ownership of the event loop)

c0deright:
Hold your beer - it seems this doesn't fix https://github.com/paritytech/parity/issues/8618. Please see https://github.com/paritytech/parity/issues/8618#issuecomment-395076883

debris:
🍺

ipc/src/select_both.rs
```diff
@@ -0,0 +1,69 @@
...
+
+impl<S1, S2> Stream for SelectBoth<S1, S2>
+	where S1: Stream,
+		  S2: Stream<Item = S1::Item, Error = S1::Error>
+{
+	type Item = S1::Item;
+	type Error = S1::Error;
+
+	fn poll(&mut self) -> Poll<Option<S1::Item>, S1::Error> {
+		let (a, b) = if self.flag {
```
rphmeier: eating a virtual dispatch on every call to poll doesn't seem great to me.

ipc/src/select_both.rs
```diff
@@ -0,0 +1,69 @@
...
+		};
+		self.flag = !self.flag;
+
+		match a.poll()? {
+			Async::Ready(Some(item)) => return Ok(Some(item).into()),
+			Async::Ready(None) => return Ok(None.into()),
+			Async::NotReady => (),
+		};
+
+		self.flag = !self.flag;
```
rphmeier: is this second negation of `self.flag` intended?

ipc/src/select_both.rs
rphmeier: (a macro `poll_inner!(a, b)` would probably end up being less verbose in the end)

ipc/src/select_both.rs
debris: yes

ipc/src/select_both.rs
debris: it is just modified the version of `select` from futures library https://docs.rs/futures/0.1.21/src/futures/stream/select.rs.html#11-15

ipc/src/server.rs
```diff
@@ -332,13 +339,15 @@ mod tests {
```
dvdplm: This test still occasionally fails here. Running `for i in {1..20}; do cargo test; done` it fails 2-3 times. Running it on by itself (`for i in {1..20}; do cargo test req_parallel; done`) makes things worse: about a fourth fails.

ipc/src/server.rs
```diff
@@ -316,58 +317,63 @@ mod tests {
...
 	#[test]
 	fn req_parallel() {
-		use std::thread;
-
 		::logger::init_log();
 		let path = "/tmp/test-ipc-45000";
 		let server = run(path);
-		thread::sleep(::std::time::Duration::from_millis(1000));
+		thread::sleep(Duration::from_millis(100));
+		let (stop_signal, stop_receiver) = mpsc::channel(400);
```
dvdplm: How about we parametrize the nr of threads/loops so we can do `mpsc::channel(TEST_THRDS * TEST_ITERS);`?

ipc/src/server.rs
```diff
@@ -316,58 +317,63 @@ mod tests {
...
-							result,
-							"{\"jsonrpc\":\"2.0\",\"result\":\"hello\",\"id\":1}",
-							"Response does not exactly match the expected response",
-							);
+						stop_signal.try_send(result).unwrap();
 					}
 				})
 			);
 		}
+		thread::sleep(Duration::from_millis(100));
```
dvdplm: Just curious: why do we still need to nap?

debris:
this issue is super annoying. it works on my machine always now. But it looks like the server is not yet ready, when the tests start

dvdplm:
> it works on my machine always now

I still get failures, but only using 1.26.1. No failures on nightly.

ipc/Cargo.toml
```diff
@@ -20,7 +20,7 @@ env_logger = "0.5"
 lazy_static = "1.0"
 
 [target.'cfg(not(windows))'.dev-dependencies]
-tokio-uds = "0.1"
+tokio-uds = "0.2"
```
debris: Updating the package helped me diagnose tests issues

ipc/src/server.rs
```diff
@@ -172,7 +172,9 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
 					})
 				})
 				.filter_map(|x| x)
-				.select(receiver.map_err(|e| {
+				// we use `select_with_weak` here, instead of `select`, to close the stream
+				// as soon as the ipc pipe is closed
+				.select_with_weak(receiver.map_err(|e| {
```
debris: maintains the stream as long as the connection exists. the outgoing stream may be closed earlier

ipc/src/server.rs
```diff
@@ -187,6 +189,7 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
 
 				Ok(())
 			});
+			start_signal.send(Ok(())).expect("Cannot fail since receiver never dropped before receiving");
```
debris: `start_signal` was triggered before the server was fully set up. After fixing the order of initialisation, we no longer need `thread::sleep` in tests

ipc/src/server.rs
tomusdrw: I don't get why we use `select_with_weak`, why not simlpy spawn the two streams separately? The point of `select` is to end the stream if any of the stream ends, now we have `select_with_weak` which "Finishes when strong stream finishes", so what's the point of using `select` at all if you can spawn the stream separately?

PR #286, opened 2018-07-03
Use serde attributes instead of custom ser/de

tomusdrw:
Couple of breaking changes:
1. `Call::Invalid` now has `id` field
2. `params:[]` is no longer deserialized to `Params::None`, but rather to `Params::Array(vec![])`
3. Stringified `Id`s are no longer parsed to numbers so `"2"` is `Id::Str("2")` not `Id::Num(2)`
4. Passing a non-object (and non-array) as a `Request` is now parse error (earlier it was deserialized to invalid call).

I suspect this should give a slight performance gain as well, but haven't benchmark it yet.

### DIFF
--- a/core/src/types/request.rs
+++ b/core/src/types/request.rs
@@ -1,9 +1,6 @@
 //! jsonrpc request
-use serde::de::{Deserialize, Deserializer, Error as DeError};
-use serde::ser::{Serialize, Serializer, Error as SerError};
-use serde_json::{from_value, Error as JsonError};
 
-use super::{Id, Params, Version, Value};
+use super::{Id, Params, Version};
 
 /// Represents jsonrpc request which is a method call.
 #[derive(Debug, PartialEq, Deserialize, Serialize)]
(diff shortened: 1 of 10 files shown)


tomusdrw:
@debris One additional breaking change:
`Option<Params>` is changed to just `Params`.

The type already contains `Params::None` and with serde we are not really able to distinguish between `None` and `Some(Params::None)` (which is IMHO useless anyway).

core/src/types/request.rs
```diff
@@ -32,18 +30,32 @@ pub struct Notification {
...
 	/// Call method
 	MethodCall(MethodCall),
 	/// Fire notification
 	Notification(Notification),
 	/// Invalid call
-	Invalid(Id),
+	Invalid {
+		/// Call id (if known)
+		#[serde(default = "default_id")]
+		id: Id,
```
debris: does it work? is there a test for it?

debris: I see there is, nvm

PR #288, opened 2018-07-04
Avoid Boxing in middleware

tomusdrw:
Complicates the types a bit, but should work better.

I hoped to use `impl Future`, but at some point we use the returned type as associated type in a trait and using `impl Future` there is not yet possible (will it ever be?).

Other approach would be to use `impl Future` and then `Box` wherever we use associated type (kind of anticipating that it's going to be faster in (near) future), but I went for the performant option for now.


tomusdrw:
@debris ping

PR #296, opened 2018-07-31
Enable using Trailing in tests.

tomusdrw:


tomusdrw:
@dvdplm I'm working on PR to polkadot that allows you to subscribe to storage changes. It has an API that uses `Trailing<Vec<StorageKey>>` (so you can either omit last parameter or provide a vector of storage keys).
Unfortunatelly currently I'm unable to create `Trailing` object in tests, since it's private for this crate. This PR adds a conversion from `Option<T>` into `Trailing<T>` so I can do `None.into()` or `Some(vec![StorageKey(vec![1])]).into()` in tests.

Issue #194, opened 2017-08-27
STDIN/STDOUT server


parisk:
Hi there,

I would like to ask a question about this project.

Is there a way to set up a "server" that reads JSON RPC requests from STDIN and prints the output to STDOUT?

I have read part of the source code, but I am not quite sure yet how it works.

**Disclaimer**: I am also totally new to Rust, so this could be total newbie question 😄.

tomusdrw:
Hey!
Currently there is no such server implemented, but the implementation should be pretty straight forward.

You can however use `jsonrpc-tcp-server` and `nc` tool:
```
$ cargo run --example tcp -p jsonrpc-tcp-server
$ echo '{"jsonrpc":"2.0","id":1,"method":"say_hello","params":[]}' | nc -v localhost 3030
Connection to localhost 3030 port [tcp/*] succeeded!
{"jsonrpc":"2.0","result":"hello","id":1}
```

bemeurer:
I'll be taking this on :) 

@tomusdrw, any tips?

tomusdrw:
@bemeurer 
Perhaps consider those libraries:

https://github.com/paulkernfeld/tokio-stdin
https://github.com/paulkernfeld/tokio-stdout

Author mentions that these are not production ready, but might be a good opportunity to contribute to that projects as well. That said I don't think that stdin/stdout jsonrpc server would ever be useful in production environment, so we don't have to worry that much.

bemeurer:
@tomusdrw I'm starting work on this right now. If it's not too much to ask, and mostly because I haven't done anything with the jsonrpc library before, could you give me some mentoring on what needs to be done?
What I mean by that is some kind of either roadmap or a north on where to start.

tomusdrw:
Hi @bemeurer, personally I like to start from an example or a test:
https://github.com/paritytech/jsonrpc/blob/master/ws/examples/ws.rs#L5
(you don't need to start with super complicated `ServerBuilder` at the beginning)

Later it would be just a matter of calling `IoHandler::handle_request` method: https://github.com/paritytech/jsonrpc/blob/master/core/src/io.rs#L169

So you only need to get a new-line delimited string from `tokio-stdin` and then pass the response to `tokio-stdout` when it's ready. Could probably be done nicely using `Stream` and `Sink` traits initially.

At the very beginning I would skip:
1. Metadata,
2. Extractors
3. Sessions & PubSub

just to make the simplest example work and later add those features in subsequent commits/PRs.
[...]

PR #299, opened 2018-08-17
Add stdio server

cmichi:
Hi @tomusdrw,

this is the current status of #194.

It's clear to me that this is most probably not the version we want to merge. I think of it more as a draft to get feedback on and since I'm new to Rust I would much appreciate any feedback!                                                                                                                                                 

It works and you can test it e.g. by

```
yes '{"jsonrpc":"2.0","id":1,"method":"say_hello","params":[]}' | head -n25 | cargo run --example stdin-stdout -p jsonrpc-stdin-stdout-server                      
```

I decided to use https://github.com/vi/tokio-stdin-stdout/ instead of the two libs you had mentioned, because it looked more elaborate and maintained to me.       

Some things:

1. There is currently no test for the server and I am unsure how to implement one. The thing is that all these stdin/stdout libs read from `::std::io::stdin();`, which I think is a problem. As I understand it Rust tests are executed in parallel and hence concurrent writes to `stdin` by different tests can occur. So different tests which use stdin would probably interfere with each other.    
<br/>I guess one option could be to spawn the whole stdin-stdout server as a separate process, but from what I have seen so far this seems a bit unusual.

2. The instantiation currently differs from the other servers, for which you use an API like this:                                                                 

```
  .start().unwrap())
  .expect("Unable to start RPC server");

server.wait().unwrap()
```

What do you think would be a good way to align the instantiation?

3. Regarding the `Make request processing log optional` commit: I'm not that happy with this solution, since it changes behavior for all the other servers too. What do you think? Maybe it also makes sense to persist  somewhere in the readme's that one needs to explicitly set the `info` flag to still get this output.                                      

Thanks!

### DIFF
--- a/stdio/README.md
+++ b/stdio/README.md
@@ -1,0 +1,32 @@
+# jsonrpc-stdio-server
+STDIN/STDOUT server for JSON-RPC 2.0.
+Takes one request per line and outputs each response on a new line.
+
+[Documentation](http://paritytech.github.io/jsonrpc/jsonrpc_stdio_server/index.html)
+
+## Example
+
+`Cargo.toml`
+
+```
+[dependencies]
+jsonrpc-stdio-server = { git = "https://github.com/paritytech/jsonrpc" }
+```
+
+`main.rs`
+
+```rust
+extern crate jsonrpc_stdio_server;
+
+use jsonrpc_stdio_server::server;
+use jsonrpc_stdio_server::jsonrpc_core::*;
+
+fn main() {
+	let mut io = IoHandler::default();
+	io.add_method("say_hello", |_params| {
+		Ok(Value::String("hello".to_owned()))
+	});
+
+	ServerBuilder::new(io).build();
+}
+```
--- a/stdio/examples/stdio.rs
+++ b/stdio/examples/stdio.rs
@@ -1,0 +1,13 @@
+extern crate jsonrpc_stdio_server;
+
+use jsonrpc_stdio_server::ServerBuilder;
+use jsonrpc_stdio_server::jsonrpc_core::*;
+
+fn main() {
+	let mut io = IoHandler::default();
+	io.add_method("say_hello", |_params| {
+		Ok(Value::String("hello".to_owned()))
+	});
+
+	ServerBuilder::new(io).build();
+}
--- a/stdio/src/lib.rs
+++ b/stdio/src/lib.rs
@@ -1,0 +1,76 @@
+//! jsonrpc server using stdin/stdout
+//!
+//! ```no_run
+//! extern crate jsonrpc_stdio_server;
+//!
+//! use jsonrpc_stdio_server::ServerBuilder;
+//! use jsonrpc_stdio_server::jsonrpc_core::*;
+//!
+//! fn main() {
+//! 	let mut io = IoHandler::default();
+//! 	io.add_method("say_hello", |_params| {
+//! 		Ok(Value::String("hello".to_owned()))
+//! 	});
+//!
+//! 	ServerBuilder::new(io).build();
+//! }
+//! ```
+
+#![warn(missing_docs)]
+
+extern crate futures;
+extern crate tokio;
+extern crate tokio_codec;
+extern crate tokio_io;
+extern crate tokio_stdin_stdout;
+#[macro_use] extern crate log;
+
+pub extern crate jsonrpc_core;
+
+use std::sync::Arc;
+use tokio::prelude::{Future, Stream};
+use tokio_codec::{FramedRead, FramedWrite, LinesCodec};
+use jsonrpc_core::{IoHandler};
+
+/// Stdio server builder
+pub struct ServerBuilder {
+	handler: Arc<IoHandler>,
+}
+
+impl ServerBuilder {
+	/// Returns a new server instance
+	pub fn new<T>(handler: T) -> Self where T: Into<IoHandler> {
+		ServerBuilder { handler: Arc::new(handler.into()) }
+	}
+
+	/// Will block until EOF is read or until an error occurs.
+	/// The server reads from STDIN line-by-line, one request is taken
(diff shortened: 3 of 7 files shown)


stdin-stdout/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "jsonrpc-stdin-stdout-server"
```
NikVolf: maybe "jsonrpc-stdio-server" would be a better name?

stdin-stdout/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+description = "STDIN/STDOUT server for JSON-RPC"
+version = "8.0.1"
```
NikVolf: 0.1 would be better starting version?

stdin-stdout/Cargo.toml
cmichi: Good point, I renamed it.

stdin-stdout/Cargo.toml
cmichi: Ah yeah, missed that on copy/paste. I have amended this PR with your remarks.

core/src/io.rs
```diff
@@ -167,6 +167,7 @@ impl<T: Metadata, S: Middleware<T>> MetaIoHandler<T, S> {
 
 	/// Handle given request asynchronously.
 	pub fn handle_request(&self, request: &str, meta: T) -> FutureResult<S::Future> {
+		info!("handle_request: {:?}", request);
```
NikVolf: it will produce output with default logging, is it desirable?

tomusdrw: I think we should remove it, we already have `trace` couple of lines below.

stdio/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+      Some(res) => res,
+      None => {
+        eprintln!("JSON RPC request produced no response: {:?}", input);
+        String::from("")
+      }
+    })
+  )
+}
+
+pub fn server(io: IoHandler) {
```
tomusdrw: Please add documentation to this function and `#![warn(missing_docs)]` to the entire crate.

stdio/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+extern crate tokio_io;
+extern crate tokio_stdin_stdout;
+
+pub extern crate jsonrpc_core;
+use jsonrpc_core::{IoHandler};
+
+use tokio::prelude::{Future, Stream};
+use tokio_codec::{FramedRead, FramedWrite, LinesCodec};
+
+fn process(io: &IoHandler, input: String) -> Box<Future<Item = String, Error = ()> + Send> {
```
tomusdrw: maybe return `impl Future<Item = String, Error =()> + Send` here to avoid boxing?

stdio/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+  Box::new(
+    io.handle_request(&input).map(move |result| match result {
+      Some(res) => res,
+      None => {
+        eprintln!("JSON RPC request produced no response: {:?}", input);
```
tomusdrw: Please use logger for that, it's also not really an error, it just means that the request was `Notification` so server was not supposed to respond anything.

stdio/examples/stdio.rs
```diff
@@ -0,0 +1,13 @@
...
+use jsonrpc_stdio_server::server;
+use jsonrpc_stdio_server::jsonrpc_core::*;
+
+fn main() {
+  let mut io = IoHandler::default();
+  io.add_method("say_hello", |_params| {
+    Ok(Value::String("hello".to_owned()))
+  });
+
+  server(io);
```
tomusdrw: I would prefer a `ServerBuilder` to be inline with other transports.
(It would be super simple though, just `SeverBuilder::new(io).build()`)

stdio/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+pub fn server(io: IoHandler) {
+  let stdin = tokio_stdin_stdout::stdin(0);
+  let stdout = tokio_stdin_stdout::stdout(0).make_sendable();
+
+  let framed_stdin = FramedRead::new(stdin, LinesCodec::new());
+  let framed_stdout = FramedWrite::new(stdout, LinesCodec::new());
+
+  let future = framed_stdin.and_then(move |line| {
```
tomusdrw: The whole block could look like this:
```rust
  let future = framed_stdin
	  .and_then(move |line| process(&io, line).map_err(|_| unreachable!()))
	  .forward(framed_stdout)
	  .map(|_| ())
	  .map_err(|e| panic!("{:?}", e));

  tokio::run(future)
```

core/src/io.rs
cmichi: I dropped the commit, so that existing behavior remains unchanged.

stdio/src/lib.rs
cmichi: @tomusdrw Thanks for the comments! I included them and amended the commit. Can you please take another look?

cmichi:
@tomusdrw Thanks for the comments! I included them and amended the commit. Can you please take another look?

stdio/README.md
```diff
@@ -0,0 +1,32 @@
...
+use jsonrpc_stdio_server::server;
+use jsonrpc_stdio_server::jsonrpc_core::*;
+
+fn main() {
+	let mut io = IoHandler::default();
+	io.add_method("say_hello", |_params| {
+		Ok(Value::String("hello".to_owned()))
+	});
+
+	server(io);
```
tomusdrw: Should be `ServerBuilder`?

cmichi:
Good catch, fixed it!

NikVolf:
Thanks!

cmichi:
Great, thank you!

PR #300, opened 2018-08-23
Allow specifying access permissions for IPC endpoint for IPC server.

pinkisemils:
Along with enabling changes from https://github.com/NikVolf/parity-tokio-ipc/pull/6, this PR enables the user of `jsonrpc_ipc_server` to create a server with a user provided `parity_tokio_ipc::Endpoint`. 
This is required so that the user can set the security attributes of the named pipe, which in our case allows us to have unprivileged processes connect to the pipe when the server is running as a system service.

Once the PR for `parity-tokio-ipc` is merged, I will change the `parity-tokio-ipc` entry in `ipc/Cargo.toml` to point back towards the original repository.

### DIFF
--- a/ipc/src/server.rs
+++ b/ipc/src/server.rs
@@ -16,6 +16,8 @@ use parking_lot::Mutex;
 
 use meta::{MetaExtractor, NoopExtractor, RequestContext};
 use select_with_weak::SelectWithWeakExt;
+use parity_tokio_ipc::Endpoint;
+pub use parity_tokio_ipc::SecurityAttributes;
 
 /// IPC server session
 pub struct Service<M: Metadata = (), S: Middleware<M> = NoopMiddleware> {
@@ -52,6 +54,7 @@ pub struct ServerBuilder<M: Metadata = (), S: Middleware<M> = NoopMiddleware> {
 	remote: reactor::UninitializedRemote,
 	incoming_separator: codecs::Separator,
 	outgoing_separator: codecs::Separator,
+	security_attributes: SecurityAttributes,
 }
 
 impl<M: Metadata + Default, S: Middleware<M>> ServerBuilder<M, S> {
@@ -76,6 +79,7 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
 			remote: reactor::UninitializedRemote::Unspawned,
 			incoming_separator: codecs::Separator::Empty,
 			outgoing_separator: codecs::Separator::default(),
+			security_attributes: SecurityAttributes::empty(),
 		}
 	}
 
@@ -106,7 +110,13 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
 		self
 	}
 
-	/// Run server (in a separate thread)
+	/// Sets the security attributes for the underlying IPC socket/pipe
+	pub fn set_security_attributes(mut self, attr: SecurityAttributes) -> Self {
+		self.security_attributes = attr;
+		self
+	}
+
+	/// Creates a new server from the given endpoint.
 	pub fn start(self, path: &str) -> std::io::Result<Server> {
 		let remote = self.remote.initialize()?;
 		let rpc_handler = self.handler;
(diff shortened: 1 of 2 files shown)


ipc/src/server.rs
```diff
@@ -104,11 +105,11 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
 		self
 	}
 
-	/// Run server (in a separate thread)
-	pub fn start(self, path: &str) -> std::io::Result<Server> {
+	/// Creates a new server from the given endpoint.
+	pub fn from_endpoint(self, endpoint: Endpoint) -> std::io::Result<Server> {
```
tomusdrw: I don't like the fact that we require some external type here.

I think we should:
1. Leave `start` as is
2. Expose `pipe_security` (or sth similar) method on the `ServerBuilder`.
3. Keep `parity-tokio-ipc` as implementation detail.

pinkisemils: That would involve keeping a `*mut SECURITY_ATTRIBUTES` field in the server builder. The whole situation around Windows ACL APIs for now is a bit messy because there isn't a nice Rust wrapper around it. `winapi` is the only thing that I'm aware of for now. If you believe that the pointer should still be stored on the server builder, I can go ahead and implement it that way!

tomusdrw: `parity-tokio-ipc` is supposed to be an abstraction for Unix and Windows sockets, in JSONRPC-IPC server we don't really care about the platform or even about particular permissions. As Niko suggested in https://github.com/NikVolf/parity-tokio-ipc/pull/6#discussion_r212315188 we should have a safe wrapper around the API anyway.

In server builder we could even have a method: `allow_unprivileged_access()` that doesn't take any parameters but just sets the required flags.

pinkisemils: Well, I've updated both `parity-tokio-ipc` and this PR. I think this one deserves to be renamed as it has nothing to do with exposing a way to construct a server from an endpoint.

My original intention was to allow to separate the server code further away from the internals of the `parity-tokio-ipc::Endpoint`, as to allow to keep the platform specific stuff inside the Endpoint. Currently, the server implementation has to be _platform aware_.

ipc/src/server.rs
```diff
@@ -219,10 +235,32 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
 			Err(e) => Err(e)
 		}
 	}
+
+	/// Creates the underlying named pipes with `READ_GEENRIC | FILE_WRITE_DATA` access permissions to Everyone.
+	#[cfg(windows)]
+	pub fn allow_everyone_to_connect(&mut self) -> ::std::io::Result<()> {
```
NikVolf: no need in such shortcuts here, better leave in `parity-tokio-ipc` only

ipc/src/server.rs
```diff
@@ -219,10 +235,32 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
...
+		self.security_attributes = Some(SecurityAttributes::allow_everyone_connect()?);
+		Ok(())
+	}
+
+	/// Creates the underlying named pipes with `READ_GEENRIC | WRITE_GENERIC` access permissions to Everyone.
+	#[cfg(windows)]
+	pub fn allow_everyone_to_create_pipes(&mut self) -> ::std::io::Result<()> {
```
NikVolf: ditto, let's leave it in parity-tokio-ipc

ipc/src/server.rs
```diff
@@ -219,10 +235,32 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
...
+		self.security_attributes = Some(SecurityAttributes::allow_everyone_create()?);
+		Ok(())
+	}
+
+	/// Creates the underlying named pipes with `READ_GEENRIC | WRITE_GENERIC` without specifying any security
+	/// attributes.
+	#[cfg(windows)]
+	pub fn default_security(&mut self) {
```
NikVolf: ditto, let's leave it in parity-tokio-ipc

pinkisemils: So, you think that `parity_tokio_ipc::SecurityAttributes` should still be exposed to the consumers of of `jsonrpc_ipc_server::ServerBuilder`? Just trying to understand the end goal here :)

pinkisemils:
I've updated both `parity-tokio-ipc` and this branch to not expose the `Endpoint` type, but rather just a safe wrapper around `SecurityAttributes`. I'm happy to change more things if required.

NikVolf:
Please rebase, @pinkisemils

NikVolf:
Awesome, thanks!

PR #301, opened 2018-08-24
Health API handling.

tomusdrw:
Allows to expose health endpoint for HTTP server that delegates to one of the RPCs.

### DIFF
--- a/http/src/handler.rs
+++ b/http/src/handler.rs
@@ -366,7 +366,7 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
 	) -> Result<RpcPollState<M, S::Future>, hyper::Error> {
 		use self::core::types::{Call, MethodCall, Version, Params, Request, Id, Output, Success, Failure};
 
-		// Parse request
+		// Create a request
 		let call = Request::Single(Call::MethodCall(MethodCall {
 			jsonrpc: Some(Version::V2),
 			method,
@@ -414,7 +414,7 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
 			params.push(v)
 		}
 
-		// Parse request
+		// Create a request
 		let call = Request::Single(Call::MethodCall(MethodCall {
 			jsonrpc: Some(Version::V2),
 			method: method.into(),
--- a/http/src/handler.rs
+++ b/http/src/handler.rs
@@ -23,6 +23,7 @@ pub struct ServerHandler<M: Metadata = (), S: Middleware<M> = NoopMiddleware> {
 	cors_max_age: Option<u32>,
 	middleware: Arc<RequestMiddleware>,
 	rest_api: RestApi,
+	health_api: Option<(String, String)>,
 	max_request_body_size: usize,
 }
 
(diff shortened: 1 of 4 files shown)


http/src/handler.rs
```diff
@@ -311,13 +343,57 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
...
 	}
 
+	fn process_health(
+		&self,
+		method: String,
+		metadata: M,
+	) -> Result<RpcPollState<M, S::Future>, hyper::Error> {
+		use self::core::types::{Call, MethodCall, Version, Params, Request, Id, Output, Success, Failure};
+
+		// Parse request
```
NikVolf: what exactly is parsed here

http/src/handler.rs
```diff
@@ -311,13 +343,57 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
...
+		let call = Request::Single(Call::MethodCall(MethodCall {
+			jsonrpc: Some(Version::V2),
+			method,
+			params: Params::None,
+			id: Id::Num(1),
```
NikVolf: can't it clash with actual requests?

http/src/handler.rs
tomusdrw: Good catch, blidndly copied a comment from `process_rest` (which was incorrect as well). Fixed.

http/src/handler.rs
tomusdrw: It doesn't really matter - `id` is only used by the client to correlate request and response. Here:
1. We are in HTTP context so we have correlation already
2. We actually have a `Future` for the response as well, so we don't need `id`
3. `jsonrpc-core` is not using `id ` for anything, it's not even passed to handlers.
4. It's not part of the response at all, so it's just internal quirk to use existing types & methods.

NikVolf: but can't client mess responses if he gets two with id="1"?

NikVolf: (one that he sent himself, one from health response)

http/src/lib.rs
```diff
@@ -261,7 +263,19 @@ impl<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>> ServerBuilder<M, S> {
...
 
-	/// Sets Enables or disables HTTP keep-alive.
+	/// Enable health endpoint.
+	///
+	/// Allows you to expose one of the methods under `GET /<path>`
+	/// The method will be invoked with no parameters.
+	/// Error returned from the method will be converted to status `500` response.
+	///
+	/// Expects a tuple with `(<path>, <rpc-method-name>)`.
+	pub fn health_api<'a, T: Into<Option<(&'a str, &'a str)>>>(mut self, health_api: T) -> Self {
```
NikVolf: why not `Into<Option<(String, String)>>` ?

tomusdrw:
@debris The idea is to have a status endpoint for HTTP server, and as the implementation we just use one of the already existing RPC methods.

The general idea is that the service that exposes RPC can be monitored with regular monitoring/load balancing tools that expect some kind of status endpoint returning HTTP 200 OK if the service is functional.
This PR makes it simple to re-use one of the RPC methods for that purpose, so you don't have to wrap the service in some 3rd party proxy.

The specific use case is to bring back `/health` endpoint after it was removed together with dapps-stuff in Parity Ethereum. Requested by users here: https://github.com/paritytech/parity-ethereum/issues/9401

PR #302, opened 2018-08-24
Http, map channel send errors to future instead of panic

CriesofCarrots:
Issue: attempting to launch a server with multiple threads and bind to a port already in use occasionally triggers a panic (SENDER_PROOF) in a server thread, in addition to the ServerBuilder AddrInUse error. Currently, the only way to handle channel send panics like these from main is a custom panic hook.

This PR propagates channel send errors to the server future, so they receive the same error handling as subsequent server operations.


NikVolf:
Thanks!

Issue #114, opened 2017-04-16
Allow customisation of Access-Control-Allow-Headers


ghost:
## Steps to reproduce:

1. Expand `Access-Control-Request-Headers: Origin, Content-Type, Accept` in request with `Authorization` HTTP header.

## What expected:

- `Access-Control-Allow-Headers` from response will contain `Authorization` as accepted HTTP header

## What happened by fact:

- `Access-Control-Allow-Headers` from response doesn't contain `Authorization` HTTP header

## Environment:

```
> rustc --version --verbose
rustc 1.18.0-nightly (bbdaad0dc 2017-04-14)
binary: rustc
commit-hash: bbdaad0dc8dc64e036ccee817f90a91876b32a9d
commit-date: 2017-04-14
host: x86_64-pc-windows-msvc
release: 1.18.0-nightly
LLVM version: 3.9
```

```
Server: jsonrpc-minihttp-server/7.0.0
```

## Related info:

- [HTTP Authentication: Basic and Digest Access Authentication](https://tools.ietf.org/html/rfc2617)

dulanov:
from Chrome's `chrome://net-internals/#events`:

```
719027: URL_REQUEST
http://localhost:1920/
Start Time: 2017-04-16 23:18:58.760

t=532210 [st=0] +REQUEST_ALIVE  [dt=7]
                 --> priority = "MEDIUM"
                 --> url = "http://localhost:1920/"
t=532210 [st=0]   +URL_REQUEST_DELEGATE  [dt=4]
t=532210 [st=0]      DELEGATE_INFO  [dt=4]
                     --> delegate_blocked_by = "Adblock Plus-Erweiterung"
t=532214 [st=4]   -URL_REQUEST_DELEGATE
t=532214 [st=4]   +URL_REQUEST_START_JOB  [dt=3]
                   --> load_flags = 34624 (DO_NOT_SAVE_COOKIES | DO_NOT_SEND_AUTH_DATA | DO_NOT_SEND_COOKIES | MAYBE_USER_GESTURE | VERIFY_EV_CERT)
                   --> method = "OPTIONS"
                   --> url = "http://localhost:1920/"
t=532214 [st=4]      URL_REQUEST_DELEGATE  [dt=0]
t=532214 [st=4]      HTTP_CACHE_GET_BACKEND  [dt=0]
t=532214 [st=4]     +HTTP_STREAM_REQUEST  [dt=1]
[...]

PR #305, opened 2018-09-03
Allow customising access control allow headers

cmichi:
This should implement #114.

I also discovered that @tomusdrw made a comment at https://github.com/paritytech/parity-ethereum/issues/6616 which seems to refer to the same issue. So I think that issue would be solved through this PR as well. I implemented this PR in the way suggested in that issue (with `Only` and `Any()` enums).

A thing that I am a bit unsure about is the set of header fields which are always allowed (`get_always_allowed_headers()` in the PR). This is something we might need to discuss. The behavior of this PR is that even when `AccessControlAllowHeaders::Only` is used there are still some fields which are always allowed (`Host` etc.).


http/src/handler.rs
```diff
@@ -33,6 +36,7 @@ impl<M: Metadata, S: Middleware<M>> ServerHandler<M, S> {
 		jsonrpc_handler: Rpc<M, S>,
 		cors_domains: CorsDomains,
 		cors_max_age: Option<u32>,
+		allowed_headers: AccessControlAllowHeaders,
```
tomusdrw: Please refer to the type in a consistent way (either `cors::AccessControlAllowHeaders` or directly). TBH I prefer the former.

http/src/handler.rs
```diff
@@ -494,16 +514,18 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
...
 				Method::Post
 			]));
-			headers.set(header::AccessControlAllowHeaders(vec![
-				Ascii::new("origin".to_owned()),
-				Ascii::new("content-type".to_owned()),
-				Ascii::new("accept".to_owned()),
-			]));
+
+			if let AllowHeaders::Ok(cors_allow_headers) = cors_allow_headers {
+				if cors_allow_headers.len() > 0 {
```
tomusdrw: `!....is_empty()`?

server-utils/src/cors.rs
```diff
@@ -201,18 +246,98 @@ pub fn get_cors_header(origin: Option<&str>, host: Option<&str>, allowed: &Optio
...
+					AllowHeaders::Ok(empty)
+				},
+			}
+		},
+		Some(requested) => {
+			// "requested" contains the fields for which it is inquired to know
+			// if they can be used.
+
+			let echo = AllowHeaders::Ok(
+				header::AccessControlAllowHeaders(requested.deref().clone())
```
tomusdrw: `.deref()` most likely unneessary

server-utils/src/cors.rs
```diff
@@ -201,18 +246,98 @@ pub fn get_cors_header(origin: Option<&str>, host: Option<&str>, allowed: &Optio
...
+
+			match cors_allow_headers {
+				AccessControlAllowHeaders::Any => {
+					// Any field is allowed. Our response are the fields about which
+					// the request inquired.
+					echo
+				},
+
+				AccessControlAllowHeaders::Only(only) => {
+					let mut allowed_headers = only.clone();
```
tomusdrw: Why do we need to clone it with every request? Maybe just do two checks instead of extending and doing one?

```rust
let are_all_allowed = requested.iter().all(|header| only.contains(header) || always_allowed.contains(header));
```

server-utils/src/cors.rs
```diff
@@ -201,18 +246,98 @@ pub fn get_cors_header(origin: Option<&str>, host: Option<&str>, allowed: &Optio
...
+
+					echo
+				}
+			}
+		}
+	}
+}
+
+/// Returns headers which are always allowed.
+pub fn get_always_allowed_headers() -> Vec<Ascii<String>> {
```
tomusdrw: Can we make it a static array? You should be able to use either `Ascii<&str>` or `Ascii<Cow<str>>`

server-utils/src/cors.rs
cmichi: I dereferenced here to extract the `vec`, if I leave it away the compiler complains about:
```
expected struct `std::vec::Vec`, found struct `cors::hyper::header::AccessControlRequestHeaders`
```

server-utils/src/cors.rs
cmichi: Good point, I went for `Vec<Ascii<&'static str>>`.

server-utils/src/cors.rs
cmichi: Ah yeah, you're right, the cloning is not necessary. I replaced it with your suggestion.

cmichi:
@tomusdrw Thanks for the feedback! I addressed your comments and amended the last commit. Can you please take a look again?

Do you have any thoughts on the default set of headers which should be returned by `get_always_allowed_headers()`?

http/examples/http_meta.rs
```diff
@@ -1,8 +1,10 @@
 extern crate jsonrpc_http_server;
+extern crate unicase;
 
-use jsonrpc_http_server::{ServerBuilder, hyper, RestApi};
+use jsonrpc_http_server::{ServerBuilder, hyper, RestApi, AccessControlAllowHeaders};
 use jsonrpc_http_server::jsonrpc_core::*;
 use self::hyper::header;
+use self::unicase::Ascii;
```
tomusdrw: IMHO we should use re-exported value - actually user doesn't really care that it's `Ascii` internally, can't we do like this:
```rust
AccessControlAllowHeaders::Only(vec![
  "Authorization",
])
```
and convert inside the builder?

server-utils/src/cors.rs
```diff
@@ -201,18 +246,99 @@ pub fn get_cors_header(origin: Option<&str>, host: Option<&str>, allowed: &Optio
...
+}
+
+/// Validates if the `AccessControlAllowedHeaders` in the request are allowed.
+pub fn get_cors_allow_headers(request_headers: &Headers, cors_allow_headers: &AccessControlAllowHeaders) -> AllowHeaders {
+	// Check if the header fields which were sent in the request are required
+	if let AccessControlAllowHeaders::Only(only) = cors_allow_headers {
+		let are_all_allowed = request_headers.iter()
+			.all(|header| {
+				only.contains(&Ascii::new(header.name().to_owned())) ||
+				get_always_allowed_headers().contains(&Ascii::new(header.name()))
```
tomusdrw: Why do we alloc `always_allowed_headers` on every iteration? Maybe just create it once and cache?

server-utils/src/cors.rs
```diff
@@ -201,18 +246,99 @@ pub fn get_cors_header(origin: Option<&str>, host: Option<&str>, allowed: &Optio
...
+	}
+}
+
+/// Validates if the `AccessControlAllowedHeaders` in the request are allowed.
+pub fn get_cors_allow_headers(request_headers: &Headers, cors_allow_headers: &AccessControlAllowHeaders) -> AllowHeaders {
+	// Check if the header fields which were sent in the request are required
+	if let AccessControlAllowHeaders::Only(only) = cors_allow_headers {
+		let are_all_allowed = request_headers.iter()
+			.all(|header| {
+				only.contains(&Ascii::new(header.name().to_owned())) ||
```
tomusdrw: Would also be good to avoid copying the header here, Can't we compare `Ascii<String>` with `Ascii<Cow>` or `Ascii<&'a str>`?

server-utils/src/cors.rs
tomusdrw: `(*requested).clone()` should work and doesn't require `Deref` to be imported. I think it's also more idiomatic.

server-utils/src/cors.rs
```diff
@@ -201,18 +246,99 @@ pub fn get_cors_header(origin: Option<&str>, host: Option<&str>, allowed: &Optio
...
+					// Any field is allowed. Our response are the fields about which
+					// the request inquired.
+					echo
+				},
+
+				AccessControlAllowHeaders::Only(only) => {
+					let are_all_allowed = requested.iter()
+						.all(|header| {
+							let name = &Ascii::new(header.as_ref());
+							only.contains(header) || get_always_allowed_headers().contains(name)
```
tomusdrw: Same as above - create once, not in every iteration.

server-utils/src/cors.rs
```diff
@@ -201,18 +246,99 @@ pub fn get_cors_header(origin: Option<&str>, host: Option<&str>, allowed: &Optio
...
+
+					echo
+				}
+			}
+		}
 	}
 }
 
+/// Returns headers which are always allowed.
+pub fn get_always_allowed_headers() -> Vec<Ascii<&'static str>> {
```
tomusdrw: I think it would be better to do a `HashSet` here instead of doing linear search every time. The list of headers looks good.

tomusdrw:
@cmichi could you rebase on latest master to fix the build?

http/examples/http_meta.rs
cmichi: Ah yeah, I agree, the API should only expose what is necessary.

I added a commit which introduces a struct `AccessControlAllowHeadersUnicase` which is a `HashSet<Ascii<&'static str>>` and is used internally now. I added an `Into` trait which converts the exposed `AccessControlAllowHeaders` to this internal type, so it's still possible to use a Vector for setting everything up.

server-utils/src/cors.rs
cmichi: I changed the type of only to `HashSet<Ascii<&'static str>>`, so it's no longer necessary to copy the header.

cmichi:
Thanks for the comments!

I have added a number of commits labelled `(…)` to the PR to make it easier reviewing them. Before merging I would squash them into the main `Support Access-Control-Request-Headers in http-server` commit though.

tomusdrw:
@cmichi Perfect, thanks!

PR #310, opened 2018-09-11
Make ipc server concurrent

### LINKED ISSUE #308: IPC server (possibly others) force requests to be processed in order.
Currently, the IPC server forces each request for a given connection to be processed in order. To mitigate this, the incoming request stream should be transformed with [`.buffer_unordrered()`](https://docs.rs/futures/0.1/futures/stream/trait.Stream.html#method.buffer_unordered). This might be the cause for #233. I've got a fix for the issue on this [branch](https://github.com/mullvad/jsonrpc/tree/make-ipc-server-concurrent), but I'm not sure whether my approach is ideal.

tomusdrw:
Closes #308 
@pinkisemils do you mind addressing the grumbles?

### DIFF
--- a/ipc/src/server.rs
+++ b/ipc/src/server.rs
@@ -6,7 +6,7 @@ use std::sync::Arc;
 use tokio_service::{self, Service as TokioService};
 use jsonrpc::futures::{future, Future, Stream, Sink};
 use jsonrpc::futures::sync::{mpsc, oneshot};
-use jsonrpc::{FutureResult, Metadata, MetaIoHandler, Middleware, NoopMiddleware};
+use jsonrpc::{Metadata, MetaIoHandler, Middleware, NoopMiddleware};
 use server_utils::{
 	tokio_codec::Framed,
 	tokio::{self, runtime::TaskExecutor, reactor::Handle},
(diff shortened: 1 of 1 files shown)


ipc/Cargo.toml
```diff
@@ -13,7 +13,7 @@ log = "0.4"
 tokio-service = "0.1"
 jsonrpc-core = { version = "8.0", path = "../core" }
 jsonrpc-server-utils = { version = "8.0", path = "../server-utils" }
-parity-tokio-ipc = { git = "https://github.com/nikvolf/parity-tokio-ipc", branch = "stable" }
+parity-tokio-ipc = { git = "https://github.com/nikvolf/parity-tokio-ipc", branch = "master" }
```
tomusdrw: We should probably use a different branch since `master` might change to often CC @NikVolf

ipc/src/server.rs
```diff
@@ -36,13 +36,25 @@ impl<M: Metadata, S: Middleware<M>> tokio_service::Service for Service<M, S> {
...
+	type Future = Box<Future<Item=Self::Response,Error=Self::Error> + Send>;
 
 	fn call(&self, req: Self::Request) -> Self::Future {
 		trace!(target: "ipc", "Received request: {}", req);
-		self.handler.handle_request(&req, self.meta.clone())
+		let fut = self.handler
+			.handle_request(&req, self.meta.clone())
+			.then(|result| {
+				match result {
+					Err(_) => {
```
tomusdrw: Maybe log a warning here. Why not propagate the error upstream?

ipc/src/server.rs
```diff
@@ -174,21 +186,11 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
...
-						Ok(None) => {
-							future::ok(None)
-						},
-						Ok(Some(response_data)) => {
-							trace!(target: "ipc", "Sent response: {}", &response_data);
-							future::ok(Some(response_data))
-						}
-					})
+				let responses = reader
+					.map(move |req| {
```
tomusdrw: Seome identation issues below

ipc/src/server.rs
```diff
@@ -174,21 +186,11 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
...
+					service.call(req)
 				})
+				.buffer_unordered(1024)
```
tomusdrw: Would be good if it was configurable.

tomusdrw:
ping @pinkisemils interested in addressing this? The approach is good, would be cool to have it configurable though.

ipc/src/server.rs
pinkisemils: It should be now. By default, I've set it to 5, I really have no clue what's a _good default_ here.

ipc/src/server.rs
pinkisemils: It's because the error type here is `()`. Should the server shut down if a handler invocation on the `MetaIoHandler` fails? If so, we can propagate here. What kind of error should it be here then?

tomusdrw: I think it's non-failing anyway, so it's good.

tomusdrw:
LGTM, @debris could you have a look?

PR #321, opened 2018-10-04
Augmented tokio::net::Incoming with Suspendable stream

seunlanlege:
fixes paritytech/parity-ethereum#9102

### DIFF
--- a/http/src/lib.rs
+++ b/http/src/lib.rs
@@ -57,7 +57,6 @@ pub use server_utils::{tokio, SuspendableStream};
 pub use handler::ServerHandler;
 pub use utils::{is_host_allowed, cors_allow_origin, cors_allow_headers};
 pub use response::Response;
-use std::time::Duration;
 
 /// Action undertaken by a middleware.
 pub enum RequestMiddlewareAction {
@@ -504,7 +503,7 @@ fn serve<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>>(
 
 			let mut http = server::conn::Http::new();
 			http.keep_alive(keep_alive);
-			let tcp_stream = SuspendableStream::new(listener.incoming(), Duration::from_secs(2));
+			let tcp_stream = SuspendableStream::new(listener.incoming());
 
 			tcp_stream
 				.for_each(move |socket| {
--- a/http/src/lib.rs
+++ b/http/src/lib.rs
@@ -53,10 +53,11 @@ use server_utils::reactor::{Executor, UninitializedExecutor};
 
 pub use server_utils::hosts::{Host, DomainsValidation};
 pub use server_utils::cors::{self, AccessControlAllowOrigin, Origin, AllowCors};
-pub use server_utils::tokio;
+pub use server_utils::{tokio, SuspendableStream};
 pub use handler::ServerHandler;
 pub use utils::{is_host_allowed, cors_allow_origin, cors_allow_headers};
 pub use response::Response;
+use std::time::Duration;
 
 /// Action undertaken by a middleware.
 pub enum RequestMiddlewareAction {
@@ -503,8 +504,9 @@ fn serve<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>>(
 
 			let mut http = server::conn::Http::new();
 			http.keep_alive(keep_alive);
+			let tcp_stream = SuspendableStream::new(listener.incoming(), Duration::from_secs(2));
 
-			listener.incoming()
+			tcp_stream
 				.for_each(move |socket| {
 					let service = ServerHandler::new(
 						jsonrpc_handler.clone(),
--- a/server-utils/src/suspendable_stream.rs
+++ b/server-utils/src/suspendable_stream.rs
@@ -30,25 +30,29 @@ impl<S, I> Stream for SuspendableStream<S>
 	type Error = ();
 
 	fn poll(&mut self) -> Result<Async<Option<Self::Item>>, ()> {
-		let mut recovered = None;
+		let mut resumed = false;
 		if let Some(mut timeout) = self.timeout.take() {
 			match timeout.poll() {
 				Ok(Async::Ready(_)) => {
-					recovered = Some(());
+					resumed = true;
 				}
 				Ok(Async::NotReady) => {
 					self.timeout = Some(timeout);
-					return Ok(Async::NotReady)
-				},
-				Err(_) => unreachable!("Polling a delay shouldn't yield any errors")
+					return Ok(Async::NotReady);
+				}
+				Err(_) => unreachable!("Polling a delay shouldn't yield any errors; qed")
 			}
 		}
 
 		loop {
 			match self.stream.poll() {
 				Ok(item) => return Ok(item),
+				Err(ref e) => if connection_error(e) {
+					warn!("Connection Error: {:?}", e);
+					continue
+				}
 				Err(err) => {
-					self.delay = if recovered.is_some() && self.delay.as_secs() < 5 {
+					self.delay = if resumed && self.delay.as_secs() < 5 {
 						self.delay * 2
 					} else {
 						self.delay
@@ -61,3 +65,10 @@ impl<S, I> Stream for SuspendableStream<S>
 		}
 	}
 }
+
+fn connection_error(e: &io::Error) -> bool {
+	e.kind() == io::ErrorKind::ConnectionRefused ||
+		e.kind() == io::ErrorKind::ConnectionAborted ||
+		e.kind() == io::ErrorKind::ConnectionReset
+}
+
--- a/server-utils/src/suspendable_stream.rs
+++ b/server-utils/src/suspendable_stream.rs
@@ -1,0 +1,70 @@
+use std::time::{Duration, Instant};
+use tokio::timer::Delay;
+use std::io;
+use tokio::prelude::*;
+
+/// A wrapper type for tokio::Incoming that stops accepting new connections
+/// for a specified amount of time once `ulimit` is reached
+pub struct SuspendableStream<S> {
+	stream: S,
+	delay: Duration,
+	timeout: Option<Delay>,
+}
+
+impl<S, I> SuspendableStream<S>
+	where S: Stream<Item=I, Error=io::Error>
+{
+	/// construct a new Suspendable stream, given tokio::Incoming
+	/// and the amount of time to pause for.
+	pub fn new(stream: S, delay: Duration) -> Self {
+		SuspendableStream {
+			stream,
+			delay,
+			timeout: None,
+		}
+	}
+}
+
+impl<S, I> Stream for SuspendableStream<S>
+	where S: Stream<Item=I, Error=io::Error>
+{
+	type Item = I;
+	type Error = ();
+
+	fn poll(&mut self) -> Result<Async<Option<Self::Item>>, ()> {
+		if let Some(ref mut timeout) = self.timeout {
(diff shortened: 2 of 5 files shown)


server-utils/src/suspendable_stream.rs
```diff
@@ -0,0 +1,70 @@
...
+
+		loop {
+			match self.stream.poll() {
+				Ok(socket) => return Ok(socket),
+				Err(err) => {
+					//  Os { code: 24, kind: Other, message: "Too many open files" }
+					if let Some(24) = err.raw_os_error() {
+						debug!("Error accepting connection: {}", err);
+						debug!("The server will stop accepting connections for {:?}", self.delay);
+						let mut timeout = Delay::new(Instant::now() + self.delay);
```
tomusdrw: I think we should instead wrap everything in another `loop` and here just do `self.timeout = Some(Delay...)` to avoid duplicated logic of `timeout` polling.

server-utils/src/suspendable_stream.rs
```diff
@@ -0,0 +1,70 @@
...
+/// A wrapper type for tokio::Incoming that stops accepting new connections
+/// for a specified amount of time once `ulimit` is reached
+pub struct SuspendableStream<S> {
+	stream: S,
+	delay: Duration,
+	timeout: Option<Delay>,
+}
+
+impl<S, I> SuspendableStream<S>
+	where S: Stream<Item=I, Error=io::Error>
```
tomusdrw: I think you could skip the bounds here, and just add `new` function for any `S` (not even a `Stream`)

server-utils/src/suspendable_stream.rs
```diff
@@ -0,0 +1,70 @@
...
+		self.timeout = None;
+
+		loop {
+			match self.stream.poll() {
+				Ok(socket) => return Ok(socket),
+				Err(err) => {
+					//  Os { code: 24, kind: Other, message: "Too many open files" }
+					if let Some(24) = err.raw_os_error() {
+						debug!("Error accepting connection: {}", err);
+						debug!("The server will stop accepting connections for {:?}", self.delay);
```
tomusdrw: I'd make it a warning. We could also consider suggesting the user to increase number of open files in their system.

http/src/lib.rs
```diff
@@ -53,10 +53,11 @@ use server_utils::reactor::{Executor, UninitializedExecutor};
 
 pub use server_utils::hosts::{Host, DomainsValidation};
 pub use server_utils::cors::{self, AccessControlAllowOrigin, Origin, AllowCors};
-pub use server_utils::tokio;
+pub use server_utils::{tokio, SuspendableStream};
 pub use handler::ServerHandler;
 pub use utils::{is_host_allowed, cors_allow_origin, cors_allow_headers};
 pub use response::Response;
+use std::time::Duration;
```
tomusdrw: Should go with other `std::` imports above

tcp/src/server.rs
```diff
@@ -10,11 +10,13 @@ use jsonrpc::futures::sync::{mpsc, oneshot};
 use server_utils::{
 	tokio_codec::Framed,
 	tokio, reactor, codecs,
+	SuspendableStream
 };
 
 use dispatch::{Dispatcher, SenderChannels, PeerMessageQueue};
 use meta::{MetaExtractor, RequestContext, NoopExtractor};
 use service::Service;
+use std::time::Duration;
```
tomusdrw: Move above to other `std::` imports.

http/src/lib.rs
```diff
@@ -503,8 +504,9 @@ fn serve<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>>(
 
 			let mut http = server::conn::Http::new();
 			http.keep_alive(keep_alive);
+			let tcp_stream = SuspendableStream::new(listener.incoming(), Duration::from_secs(2));
```
tomusdrw: 2 seconds might be a bit too much. What do you think about making the delay exponential? I.e. with every delay followed by another failed attempt we multiply the delay by 2. Thanks to this we could start with some really low values (like 5-10ms and increase dynamically if there are no open files available).

server-utils/src/suspendable_stream.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+impl<S, I> Stream for SuspendableStream<S>
+	where S: Stream<Item=I, Error=io::Error>
+{
+	type Item = I;
+	type Error = ();
+
+	fn poll(&mut self) -> Result<Async<Option<Self::Item>>, ()> {
+		let mut recovered = None;
```
dvdplm: Maybe call this `resumed` – we don't know if we've recovered at this point, we just know we're ready to try again, right?

server-utils/src/suspendable_stream.rs
```diff
@@ -0,0 +1,63 @@
...
+					return Ok(Async::NotReady)
+				},
+				Err(_) => unreachable!("Polling a delay shouldn't yield any errors")
+			}
+		}
+
+		loop {
+			match self.stream.poll() {
+				Ok(item) => return Ok(item),
+				Err(err) => {
```
dvdplm: I wonder if we should `match` on the underlying `stream`'s possible errors here and decide which ones we should suspend for and which ones are fatal? I'm not sure about this but aren't there other error conditions where we do not want to suspend-resume forever?

bkchr: Maybe you look at the following commit: https://github.com/hyperium/hyper/commit/68458cde57a20f4b3c9c306eaf9801189262e0a6
That introduced `sleep_on_error` in hyper. That's similar to what you are want to achieve here. 
Look into the `connection_error` function, this defines errors that should not end in sleeping.

server-utils/src/suspendable_stream.rs
bkchr: And that can just be an boolean, does not need to be an Option.

server-utils/src/suspendable_stream.rs
```diff
@@ -0,0 +1,63 @@
...
+		if let Some(mut timeout) = self.timeout.take() {
+			match timeout.poll() {
+				Ok(Async::Ready(_)) => {
+					recovered = Some(());
+				}
+				Ok(Async::NotReady) => {
+					self.timeout = Some(timeout);
+					return Ok(Async::NotReady)
+				},
+				Err(_) => unreachable!("Polling a delay shouldn't yield any errors")
```
bkchr: You should handle the error gracefully, looking into the documentation for tokio timer (https://docs.rs/tokio-timer/0.2.7/tokio_timer/struct.Error.html). You see that you can get an `at_capacity` error. As you are already in the situation, that you can not accept anymore connections, it can fairly easy be the point that you are also at the timer maximum capacity.

seunlanlege: Actually the only possible error is a shutdown error, taken directly from the source code.
and it occurs if the timer instance was dropped. That's not possible here.

screenshot from 2018-10-05 09-11-25

http/src/lib.rs
```diff
@@ -503,8 +503,9 @@ fn serve<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>>(
 
 			let mut http = server::conn::Http::new();
 			http.keep_alive(keep_alive);
+			let tcp_stream = SuspendableStream::new(listener.incoming());
```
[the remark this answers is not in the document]
tomusdrw: @debris 
1. `incoming` is a stream of incoming sockets
2. Disconnecting has nothing to do with that stream at all
3. Polling the stream may return a temporary `io::Error` (for instance if we can't open the connection because of "too many open files" limit)
4. we use `for_each` combinator which:
4.1. Runs for every `Ok(socket)`
4.2. Stops on the FIRST `Err()`
5. So any temporary `io::Error` will cause the entire server to terminate.
6. The stuff that @seunlanlege is working on causes temporary `io::Error` to sleep a little bit and re-try instead of terminating the server.
7. So the timeout introduced here only affects the server in case of heavy usage conditions, without the timeout the connection would be unsuccesful anyway.

server-utils/src/suspendable_stream.rs
```diff
@@ -0,0 +1,74 @@
...
+	timeout: Option<Delay>,
+}
+
+impl<S> SuspendableStream<S> {
+	/// construct a new Suspendable stream, given tokio::Incoming
+	/// and the amount of time to pause for.
+	pub fn new(stream: S) -> Self {
+		SuspendableStream {
+			stream,
+			delay: Duration::from_millis(500),
```
tomusdrw: IMHO it's way too large as the initial value. I'd go for something like 10-20ms

server-utils/src/suspendable_stream.rs
bkchr: It is a bit tricky and, but if you look into to the code here: https://github.com/tokio-rs/tokio/blob/1879bc49ce3126af4b561bfe535cbec9f7ae5b90/tokio-timer/src/timer/entry.rs#L161
You will find functions e.g. `increment` that internally returns an `at_capacity` error. To the outside you will just get a `shutdown` error, as you already mentioned. That is somehow a design error of tokio timer.
(`register_with` is called by `register` in `Delay`.)

http/src/lib.rs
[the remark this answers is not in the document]
seunlanlege: The reason we need this is because `Stream::for_each` returns a future that will end, if the underlying stream encounters an error. So far the only error we've seen in the wild is an io error of too many open files. This is a recoverable error and shouldn't terminate the server. which is what currently happens as reported here paritytech/parity-ethereum#9102.

The solution is to wait for a while before accepting any new connections, so we have the chance to complete and drop the existing connections.

server-utils/src/suspendable_stream.rs
```diff
@@ -0,0 +1,74 @@
...
+		loop {
+			match self.stream.poll() {
+				Ok(item) => return Ok(item),
+				Err(ref e) => if connection_error(e) {
+					warn!("Connection Error: {:?}", e);
+					continue
+				}
+				Err(err) => {
+					self.delay = if resumed && self.delay.as_secs() < 5 {
+						self.delay * 2
```
tomusdrw: This is only half-correct. After succesfull attempt we should actually reset the delay to the initial value (I'd say when handling `Ok(`).
Also not sure if we actually need `resumed` flag:
1. just increase the delay by `2` on every error (you can have it capped as now, but extract that to a constant)
2. In case of `Ok(item)` in the stream - reset the delay to initial value.

http/src/lib.rs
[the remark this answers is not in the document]
debris: @tomusdrw @seunlanlege can you add 1, 3, 4, 5 and 6 as the comment to the code? It will greatly help future maintainers of this code :)

server-utils/src/suspendable_stream.rs
```diff
@@ -0,0 +1,88 @@
...
+
+	fn poll(&mut self) -> Result<Async<Option<Self::Item>>, ()> {
+		if let Some(mut timeout) = self.timeout.take() {
+			match timeout.poll() {
+				Ok(Async::Ready(_)) => {}
+				Ok(Async::NotReady) => {
+					self.timeout = Some(timeout);
+					return Ok(Async::NotReady);
+				}
+				Err(_) => unreachable!("Polling a delay shouldn't yield any errors; qed")
```
bkchr: That is not correct, please look into my last comment here: https://github.com/paritytech/jsonrpc/pull/321#discussion_r222928243

PR #326, opened 2018-10-09
Backport suspendable stream into parity-1.11

seunlanlege:


http/src/lib.rs
```diff
@@ -438,23 +438,24 @@ fn serve<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>>(
...
 			let allowed_hosts = server_utils::hosts::update(allowed_hosts, &local_addr);
-
+			
 			let http = {
-				let mut http = server::Http::new();
-				http.keep_alive(keep_alive);
-				http.sleep_on_errors(true);
-				http
+				let mut h = server::Http::new();
+				h.keep_alive(keep_alive);
```
tomusdrw: can we get `sleep_on_errors` back? I believe parity-1.11 doesn't have the latest hyper which has this enabled by default.

server-utils/src/suspendable_stream.rs
```diff
@@ -0,0 +1,92 @@
+use std::time::{Duration, Instant};
+use tokio_timer::Delay;
+use std::io;
+use core::futures::prelude::*;
+use core::futures::task;
+/// `Incoming` is a stream of incoming sockets
+/// Polling the stream may return a temporary io::Error (for instance if we can't open the connection because of "too many open files" limit)
```
dvdplm: Please wrap this line.

server-utils/src/suspendable_stream.rs
```diff
@@ -0,0 +1,92 @@
...
+				Ok(item) => {
+					if self.next_delay > self.initial_delay {
+						self.next_delay = self.initial_delay;
+					}
+					return Ok(item)
+				}
+				Err(ref err) => {
+					if connection_error(err) {
+						warn!("Connection Error: {:?}", err);
+						continue
```
dvdplm: Now that I look at this again, I wonder if it's really ok to `continue` indefinitely here. If a peer returns e.g. `ConnectionRefused` wouldn't we keep hammering it in an infinite loop?

seunlanlege: Actually, this is a server, so `ConnectionError`, would mean there was an error when we tried to accept a connection, that socket could be dropped and we would try to accept a new/different connection.

seunlanlege:
@dvdplm I've added the tests, @tomusdrw looks like `parity-1.11` is still using tokio_core and not the new tokio runtime so i had to use `tokio_timerv0.1` and not `v0.2`

seunlanlege:
I've been building on nightly this whole time :see_no_evil:

PR #331, opened 2018-10-24
Http CloseHandle for closing while waiting

carllin:
Add CloseHandle to Json Http server to allow a different thread to close() after calling wait() on main server thread. Based on similar work for websockets: https://github.com/paritytech/jsonrpc/pull/151/files

### DIFF
--- a/http/src/lib.rs
+++ b/http/src/lib.rs
@@ -40,7 +40,7 @@ mod utils;
 mod tests;
 
 use std::io;
-use std::sync::{mpsc, Arc};
+use std::sync::{mpsc, Arc, Mutex};
 use std::net::SocketAddr;
 use std::thread;
 
@@ -430,7 +430,7 @@ impl<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>> ServerBuilder<M, S> {
 		Ok(Server {
 			address: local_addr?,
 			executor: Some(executors),
-			close: Some(close),
+			close: Arc::new(Mutex::new(Some(close))),
 		})
 	}
 }
@@ -554,7 +554,7 @@ fn configure_port(_reuse: bool, _tcp: &net2::TcpBuilder) -> io::Result<()> {
 pub struct Server {
 	address: SocketAddr,
 	executor: Option<Vec<Executor>>,
-	close: Option<Vec<oneshot::Sender<()>>>,
+	close: Arc<Mutex<Option<Vec<oneshot::Sender<()>>>>>,
 }
 
 const PROOF: &'static str = "Server is always Some until self is consumed.";
@@ -566,13 +566,7 @@ impl Server {
 
 	/// Closes the server.
 	pub fn close(mut self) {
-		for close in self.close.take().expect(PROOF) {
-			let _ = close.send(());
-		}
-
-		for executor in self.executor.take().expect(PROOF) {
-			executor.close();
-		}
+		self.close_handle().close();
 	}
 
 	/// Will block, waiting for the server to finish.
@@ -581,12 +575,46 @@ impl Server {
 			executor.wait();
 		}
 	}
+
+	/// Returns a handle to the server that can be used to close it while another thread is
+    /// blocking in `wait`.
+    pub fn close_handle(&mut self) -> CloseHandle {
+		let executor_close: Option<Vec<_>> = self.executor.as_mut().map(|executors| {
+			executors.iter_mut().map(|executor| executor.close_handle()).collect()
+		});
+
+        CloseHandle {
+            executor_close,
+            close: self.close.clone(),
+        }
+    }
 }
 
 impl Drop for Server {
 	fn drop(&mut self) {
-		self.executor.take().map(|executors| {
-			for executor in executors { executor.close(); }
-		});
+		self.close_handle().close();
 	}
+}
+
+/// A handle that allows closing of a server even if it owned by a thread blocked in `wait`.
+pub struct CloseHandle {
+    close: Arc<Mutex<Option<Vec<oneshot::Sender<()>>>>>,
+    executor_close: Option<Vec<Option<futures::Complete<()>>>>,
+}
+
+impl CloseHandle {
+    /// Closes the `Server`.
+    pub fn close(mut self) {
+		self.close.lock().unwrap().take().map(|close_vector| {
+			for close in close_vector {
+				let _ = close.send(());
+			}
+		});
+
+        self.executor_close.take().map(|executor_close_vector| {
+			for e in executor_close_vector {
+            	e.map(|v| v.send(()));
+        	}
+		});
+    }
 }
--- a/http/src/lib.rs
+++ b/http/src/lib.rs
@@ -577,17 +577,17 @@ impl Server {
 	}
 
 	/// Returns a handle to the server that can be used to close it while another thread is
-    /// blocking in `wait`.
-    pub fn close_handle(&mut self) -> CloseHandle {
+	/// blocking in `wait`.
+	pub fn close_handle(&mut self) -> CloseHandle {
 		let executor_close: Option<Vec<_>> = self.executor.as_mut().map(|executors| {
 			executors.iter_mut().map(|executor| executor.close_handle()).collect()
 		});
 
-        CloseHandle {
-            executor_close,
-            close: self.close.clone(),
-        }
-    }
+		CloseHandle {
+			executor_close,
+			close: self.close.clone(),
+		}
+	}
 }
 
 impl Drop for Server {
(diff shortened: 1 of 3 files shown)


http/src/lib.rs
```diff
@@ -581,12 +575,46 @@ impl Server {
 			executor.wait();
 		}
 	}
+
+	/// Returns a handle to the server that can be used to close it while another thread is
+    /// blocking in `wait`.
```
bkchr: Please use tabs here and the lines below.

http/src/lib.rs
```diff
@@ -581,12 +575,46 @@ impl Server {
...
-		self.executor.take().map(|executors| {
-			for executor in executors { executor.close(); }
-		});
+		self.close_handle().close();
 	}
 }
+
+/// A handle that allows closing of a server even if it owned by a thread blocked in `wait`.
+pub struct CloseHandle {
+    close: Arc<Mutex<Option<Vec<oneshot::Sender<()>>>>>,
```
bkchr: More whitespaces.

http/src/lib.rs
```diff
@@ -581,12 +575,46 @@ impl Server {
...
 	fn drop(&mut self) {
-		self.executor.take().map(|executors| {
-			for executor in executors { executor.close(); }
-		});
+		self.close_handle().close();
 	}
 }
+
+/// A handle that allows closing of a server even if it owned by a thread blocked in `wait`.
+pub struct CloseHandle {
```
bkchr: I think we should implement/derive `Clone` for the `CloseHandle`.

server-utils/src/reactor.rs
```diff
@@ -62,6 +62,14 @@ impl Executor {
 		self.executor().spawn(future)
 	}
 
+	pub fn close_handle(&mut self) -> Option<futures::Complete<()>> {
```
tomusdrw: Please add docs

server-utils/src/reactor.rs
```diff
@@ -163,4 +171,8 @@ impl RpcEventLoop {
 			warn!("Event Loop is already finished. {:?}", e);
 		});
 	}
+
+	pub fn close_handle(&mut self) -> Option<futures::Complete<()>>{
```
tomusdrw: Why do we need `close_handle` in executor?

server-utils/src/reactor.rs
tomusdrw: Why do we actually need this?

carllin:
I added the close_handle in Executor and RpcEventLoop because I wanted to support the following behavior:

1) Grab the CloseHandle for the Server
2) Call Wait() on the Server
3) Call CloseHandle.Close()

Unfortunately, if the CloseHandle just held a reference to a lock to the Executors (as was done in the websockets implementation), b/c the Wait() function also simultaneously needs access to the Executors, this won't work. Thus the close_handle() on the reactor and executor return the inner Complete() signaler that executor->close() would have called anyway, so now the CloseHandle no longer needs shared access/lock to the Executor.

http/src/lib.rs
carllin: futures::Sender, doesn't implement Clone(), so how should we approach this?

bkchr: Do you mean the `executor_close` can not be clonded?

tomusdrw: @carllin it's wrapped in `Arc<Mutex` anyway, so it doesn't really matter. If derive doesn't work then you need to implement `Clone` manually (and just clone the `Arc`)

carllin: @bkchr, @tomusdrw, yes the executor_close cannot be cloned.

carllin: @tomusdrw, in the current WS implementation, close() also closes the executor: https://github.com/paritytech/jsonrpc/blob/master/ws/src/server.rs#L149

carllin:
@tomusdrw, @bkchr, pinging for updates.

http/src/lib.rs
```diff
@@ -581,12 +575,50 @@ impl Server {
 			executor.wait();
 		}
 	}
+
+	/// Returns a handle to the server that can be used to close it while another thread is
+	/// blocking in `wait`.
+	pub fn close_handle(&mut self) -> CloseHandle {
```
tomusdrw: The function shouldn't mutate anything, so it should be `&self`.

```suggestion
	pub fn close_handle(&self) -> CloseHandle {
	  let executor_close: Option<Vec<_>> = self.executor.as_ref().map(|executors| {
			executors
				.iter()
				.map(|executor| executor.close_handle())
				.collect()
		});
		...
	}
```

server-utils/src/reactor.rs
```diff
@@ -62,6 +62,15 @@ impl Executor {
 		self.executor().spawn(future)
 	}
 
+	/// Returns a handle to close the underlying event loop
+	pub fn close_handle(&mut self) -> Option<futures::Complete<()>> {
```
tomusdrw: `mut` not needed
```suggestion
	pub fn close_handle(&self) -> Option<futures::Complete<()>> {
```

server-utils/src/reactor.rs
```diff
@@ -163,4 +172,10 @@ impl RpcEventLoop {
 			warn!("Event Loop is already finished. {:?}", e);
 		});
 	}
+
+	/// Returns the close signal that can be used to close the event loop even while
+	/// another thread is blocked on the event loop in "wait"
+	pub fn close_handle(&mut self) -> Option<futures::Complete<()>>{
```
tomusdrw: This function doesn't make much sense - the whole point of handle is to be able to have multiple instance of it. 
This function here will `take()` the close handle and only return it for the first caller. Subsequent calls will always yield `None`.
What's more it violates the assumption in the rest of this struct (see `expect("..")` calls) - so every other function like `close()` or `wait()` will actually panic.

I still believe that having this function is completely unnecessary. See how it's done in `WebSockets` server - we own `Executor` via `Arc<Mutex<Option<Executor>>>` and close it completely inside `CloseHandle` (actually it's not perfect there, cause cloning the handle and calling `close()` on a clone would panic, but it's a separate issue).

http/src/lib.rs
```diff
@@ -554,7 +554,7 @@ fn configure_port(_reuse: bool, _tcp: &net2::TcpBuilder) -> io::Result<()> {
 pub struct Server {
 	address: SocketAddr,
 	executor: Option<Vec<Executor>>,
```
tomusdrw: executor should most likely be wrapped in `Arc<Mutex<` as wel, so that it can be passed to `CloseHandle. Actually it would be simpler to store `CloseHandle` directly in `Server` and just clone it when requested. `CloseHandle` should handle the case if `Options` are `None`.
To simplify it further you can just have:
```rust

pub struct Server {
   address: SocketAddr,
  close_handle: CloseHandle,
}

impl Server {
  ...
  pub fn close_handle(&self) -> CloseHandle {
    self.close_handle.clone()
  }
}

#[derive(Clone)]
pub struct CloseHandle {
  handle: Arc<Mutex<Option<(Executor, Vec<oneshot::Sender<()>)>>>,
}

impl CloseHandle {
  ///  <docs ...>
  /// returns `false` if the server was closed earlier
  pub fn close(self) -> bool {
     if let Some((executor, senders)) = self.handle.lock().take() {
        executor.close();
        for s in senders { s.send(()).expect(...) } 
        true
     } else { false }
  }
}
```

server-utils/src/reactor.rs
carllin: The WebSockets server has a separate handle that it joins on in wait(), whereas the wait() implementation in this module requires access to the executor, so we wouldn't be able to call close() on the close handle while another thread is blocked on wait(). Is there a way to construct a similar, separate handle as exists in the WebSockets implementation?

carllin:
@tomusdrw The WebSockets server has a separate handle that it joins on in wait(), whereas the wait() implementation in this module requires access to the executor, so we wouldn't be able to call close() on the close handle while another thread is blocked on wait(). Is there a way to construct a similar, separate handle as exists in the WebSockets implementation?

tomusdrw:
@carllin Indeed WebSockets server runs an extra thread for itself, and in close handle we wait for that thread to finish. With HTTP the proper solution would be to add another `oneshot` future that is triggered after the serve is finished here: https://github.com/paritytech/jsonrpc/blob/2ed142d55662155b0dc76b9b5d2732300f265dd6/http/src/lib.rs#L461
And `Server::wait()` should actually wait for this, not the entire excutor to be closed.

carllin:
Hey Tomas,
     Sorry I haven't been able to get to it, I've been swamped with work at
work xD. Hopefully I might have some time closer to vacation time. Thanks
for following up!

Best,
Carl

dvdplm:
@carllin any chance we can get the changes in? If things have piled up on your end we can perhaps take over and finish it up, just let us know! :)

carllin:
@dvdplm, sorry don't have much bandwidth nowadays xD. Of somebody could drag this to the finish line that would be highly appreciated :D. Thanks for all the help!

http/src/lib.rs
dvdplm: Not sure how this would work: once the server starts waiting it has to take the lock on the `handle`, so `close()` as written above will hang on the lock?

dvdplm:
Closing in favour of #437

PR #333, opened 2018-10-30
Handle keep-alive correctly

tomusdrw:
It seems that hyper setting only affects the underlying TPC stream, but doesn't add any headers to the response to inform the client.

This PR changes the way we handle `keep_alive` in the builder, so that correct `Connection` headers are added to the response.

### DIFF
--- a/http/src/utils.rs
+++ b/http/src/utils.rs
@@ -55,15 +55,19 @@ pub fn cors_allow_headers(
 
 /// Returns an optional value of `Connection` header that should be included in the response.
 /// The second parameter defines if server is configured with keep-alive option.
-/// Return value of `Some(true)` indicates that `Connection: keep-alive` should be included,
-/// `Some(false)` indicates `Connection: close` and `None` doesn't include the header at all.
+/// Return value of `true` indicates that no `Connection` header should be returned,
+/// `false` indicates `Connection: close`.
 pub fn keep_alive(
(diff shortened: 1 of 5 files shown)


http/src/utils.rs
```diff
@@ -52,3 +52,18 @@ pub fn cors_allow_headers(
...
+/// Return value of `Some(true)` indicates that `Connection: keep-alive` should be included,
+/// `Some(false)` indicates `Connection: close` and `None` doesn't include the header at all.
+pub fn keep_alive(
+	request: &hyper::Request<hyper::Body>,
+	keep_alive: bool,
+) -> Option<bool> {
+	read_header(request, "connection")
+		.map(|val| match (keep_alive, val) {
+			(true, "keep-alive") => true,
+			_ => false,
```
debris: what are the other possible values of the `connection`?

tomusdrw: For the request/client it can only be `close` (the server may respond with `Connection: upgrade` also, but it's used for instance when upgrading from HTTP to WebSockets)

tomusdrw:
@twittner the header is only added if the client added `Connection: keep-alive` to the request, so pretty much to support HTTP/1.0. However I've noticed that although it's assumed implicitly, HTTP/1.1 clients still add the header so we will add it to the response as well.
Do you think it's superfluous?

PR #337, opened 2018-11-30
test: add make_request fn that returns value without pretty printing

azban:
was having issues with testing when trying to test against the expected serialized result because this lib only returns pretty printed. i kept the current interface the same and added a new function.

### DIFF
--- a/test/src/lib.rs
+++ b/test/src/lib.rs
@@ -86,8 +86,14 @@ impl Rpc {
 		io.into()
 	}
 
-	/// Perform a single, synchronous method call.
-	pub fn request<T>(&self, method: &str, params: &T) -> String where
+	fn request_helper<T>(
+		&self,
+		method: &str,
+		params: &T,
+		success_formatter: fn(&serde_json::Value) -> serde_json::Result<String>,
+		error_formatter: fn(&rpc::Error) -> serde_json::Result<String>,
+	) -> String
+	where
 		T: serde::Serialize,
 	{
 		use self::rpc::types::response;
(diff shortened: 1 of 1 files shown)


test/src/lib.rs
```diff
@@ -104,15 +110,32 @@ impl Rpc {
...
+	pub fn request<T>(&self, method: &str, params: &T) -> String where
+		T: serde::Serialize,
+	{
+		let success_formatter = serde_json::to_string_pretty;
+		let error_formatter = serde_json::to_string_pretty;
+		self.request_helper(method, params, success_formatter, error_formatter)
+	}
+
+	/// Perform a single, synchronous method call.
+	pub fn make_request<T>(&self, method: &str, params: &T) -> String where
```
tomusdrw: I don't really like the name that much, but don't have a single better suggestion. Maybe:
1. `request_compact(`?
2. `make_request(&str, &T, Encoding::Compact)` - `enum Encoding` would decide on the type of encoding.

tomusdrw:
@azban Ping, any chance you could address the suggestion?

azban:
i ended up testing in a different manner and don't need this change.. i can make the change if you think it would be useful, or we can close this for now.

tomusdrw:
@azban yeah, I think it might be useful. I think I'm more inclined to `make_request(&str, &T, Encoding::Pretty)` API, with `Encoding::Pretty` and `Encoding::Compact` variants.
It would be awesome if you'd update the PR, if not I can fork it and finish myself.

azban:
done.. good suggestion on enum. definitely cleaner.

PR #339, opened 2018-12-06
Bump to newest ws-rs & add max_payload

tomusdrw:


ws/src/server_builder.rs
```diff
@@ -49,6 +50,7 @@ impl<M: core::Metadata, S: core::Middleware<M>> ServerBuilder<M, S> {
 			session_stats: None,
 			executor: UninitializedExecutor::Unspawned,
 			max_connections: 100,
+			max_payload_bytes: 5 * 1024 * 1024,
```
jacogr: Cool. It will take some time for us to get there. (Think I'll try to align with the Polkadot-js UI side as well on the sizes - mostly applicable to WASM code & contracts)

tomusdrw: @jacogr that's just a default value, I'll configure this to 15MB on the substrate side.

PR #340, opened 2018-12-06
Use procedural macros 

### LINKED ISSUE #66: Use procedural macros

ascjones:
Closes #66.

Replaces the `build_rpc_trait!` macro with a trait level `#[rpc]` attribute. e.g.

```
#[rpc]
pub trait Rpc {
	/// Returns a protocol version
	#[rpc(name = "protocolVersion")]
	fn protocol_version(&self) -> Result<String>;

	/// Adds two numbers and returns a result
	#[rpc(name = "add", alias("callAsyncMetaAlias"))]
	fn add(&self, _: u64, _: u64) -> Result<u64>;

	/// Performs asynchronous operation
	#[rpc(name = "callAsync")]
	fn call(&self, _: u64) -> FutureResult<String, Error>;
}
```
and pubsub:
```
#[rpc]
pub trait Rpc {
	type Metadata;

	/// Hello subscription
	#[pubsub(subscription = "hello", subscribe, name = "hello_subscribe", alias("hello_sub"))]
	fn subscribe(&self, _: Self::Metadata, _: typed::Subscriber<String>, _: u64);

	/// Unsubscribe from hello subscription.
	#[pubsub(subscription = "hello", unsubscribe, name = "hello_unsubscribe")]
	fn unsubscribe(&self, _: Option<Self::Metadata>, _: SubscriptionId) -> Result<bool>;
}
```
Example of converting [parity-ethereum](https://github.com/paritytech/parity-ethereum/tree/aj-jsonrpc-derive).

### Tasks

- [x] Move public helper types currently in `jsonrpc-macros`, need to figure out where they should go
  - [x] auto_args wrappers: can be done inline
  - [x] delegates: move to core for now
  - [x] pubsub: move to pubsub
  - [x] `Trailing` replace with `Option` in last position
- [x] Optional metadata in unsubscribe https://github.com/paritytech/jsonrpc/pull/352
- [x] Copy tests from `jsonrpc-macros` and make them work
- [x] Fix bug with single parameter as per: https://github.com/paritytech/jsonrpc/commit/00056d96c4a4a99ea4b94118fcf14fd2ccd8c8b2#diff-61e49682a8be86b7288c62de1399069fR91
- [x] Fix compilation error with single trailing arg
- [x] Test with existing projects:
  - [x] `parity-ethereum`
  - [x] `substrate` 
- [x] Deprecate (rather than replace/remove) `jsonrpc-macros` with proc macros implementation currently in `jsonrpc-derive`. Need to find out how best to deprecate the macro or the entire crate.
- [x] Rust 2018 compatibility (fn arg names and use `crate`?)
- [x] Documentation!

### Future Improvements

- Auto convert method names from snake case e.g. `protocol_version` -> `protocolVersion`, allow overrides as per serde.
- Improve error messages with `proc-macro-diagnostics` e.g. https://github.com/dtolnay/syn/blob/master/examples/lazy-static/lazy-static/src/lib.rs which is currently a `nightly` feature. Possible to bring it in already with ourt own `nightly' features with shims as per [diesel](https://github.com/diesel-rs/diesel/blob/2ce0e4ea0fda474459139042247512f0c8b254cf/diesel_derives/src/diagnostic_shim.rs)
- Handle multiple trailing args, (already implemented here: https://github.com/paritytech/jsonrpc/compare/pzec_dependency_multiple_trailing_args)


ascjones:
Still need to do some more testing, documentation, refining etc on this - but just need some early feedback on the api. The approach I've taken is with valid rust trait syntax in order to use `syn` parsing. 

Check out the [examples](https://github.com/paritytech/jsonrpc/tree/b2d928d725b713373afbb2fdd682e47a5d165266/derive/examples), and also I've [experimented](https://github.com/paritytech/substrate/tree/aj-test-jsonrpc-derive/core/rpc) with converting substrate rpcs.

Things to note:
- `Trailing` type has been removed, any `Option` in last position is treated as an optional trailing parameter
- Pubsub methods separated into their own trait
  - this does add some ugliness in substrate since we have to repeat the mass of type constraints
- Helper types e.g. `IoDelegate` and typed `Subscriber` moved into `core` and `substrate` crates.

Feedback on the above and any suggestions appreciated.

svyatonik:
@ascjones Probably you know it, but it fails to parse args for RPCs that accept single argument - the example of failing test is [here](https://github.com/paritytech/jsonrpc/commit/00056d96c4a4a99ea4b94118fcf14fd2ccd8c8b2#diff-61e49682a8be86b7288c62de1399069fR91). The reason is that `Params::parse` fails to parse `Params::Array` of single element into single-element tuple. I've workaround for this [here](https://github.com/paritytech/jsonrpc/commit/1691c6e52b2af1d56e8c734312cb0a5d6826e857), but probably there's a better way to fix it

ascjones:
@svyatonik Actually no I hadn't discovered that issue yet, thanks for the thorough testing! I won't get back to this until Monday again - but will try to fix it properly or incorporate your workaround. Hopefully get this PR in next week.

derive/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "jsonrpc-derive"
+version = "9.0.0"
+authors = ["Andrew Jones <ascjones@gmail.com>"]
```
tomusdrw: Please change to Parity Technologies

derive/examples/generic-trait-bounds.rs
```diff
@@ -0,0 +1,65 @@
+extern crate serde;
```
tomusdrw: Have you considered `edition = 2018`?

derive/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+extern crate syn;
+
+use proc_macro::TokenStream;
+use syn::parse_macro_input;
+
+mod rpc_attr;
+mod rpc_trait;
+mod to_delegate;
+
+// todo: [AJ] docs
```
tomusdrw: I would add `#![warn(missing_docs)]` on the crate level.

derive/src/rpc_attr.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl RpcTraitAttribute {
+	pub fn try_from_trait_attribute(args: &syn::AttributeArgs) -> Result<RpcTraitAttribute, String> {
+		let mut visitor = RpcAttributeVisitor::default();
+		for nested_meta in args {
+			visitor.visit_nested_meta(nested_meta);
+		}
+
+		if visitor.meta_words.contains(&"pubsub".to_string()) {
```
tomusdrw: could we move all magic strings to constants?

derive/src/rpc_attr.rs
```diff
@@ -0,0 +1,109 @@
...
+			if let Some(name) = visitor.name {
+				Ok(RpcTraitAttribute::PubSubTrait { name })
+			} else {
+				Err("rpc pubsub trait attribute should have a name".to_string())
```
tomusdrw: `s/.to_string()/.into()/`

derive/tests/pubsub-macros.rs
```diff
@@ -0,0 +1,81 @@
...
+pub enum MyError {}
+impl From<MyError> for jsonrpc_core::Error {
+	fn from(_e: MyError) -> Self {
+		unreachable!()
+	}
+}
+
+type Result<T> = ::std::result::Result<T, MyError>;
+
+#[rpc(pubsub, name = "hello")]
```
tomusdrw: Does this mean that each pubsub now has to be a separate trait?

ascjones: Yes at the moment it does. It was a way to keep the grouping of the sub/unsub method pair which is done with a `{ }` block in the existing macro. I'm not 100% happy with it (not least beause in substrate it means copying generic soup for the separate traits). Any better suggestions to group the sub/unsub pair in the same trait?

tomusdrw: Yeah, I'm not happy with the block either maybe let's repeat the subscription name?

Alternatively let's have only one method per subscription and just pass an enum:
```
enum Subscription<SinkValue, ParamsTuple> {
  Subscribe { meta, subscriber, params },
  Unsubscribe { meta, subscriber, params },
}
```
What do you think? That might complicate the attribute significanlty (cause you need to provide name, subscribe, unsubscribe and potential aliases)

ascjones: Repeating the subscription name seems like an easy solution. However I hadn't thought of the enum idea - I will at least try it out to see how it looks in practice. Job for next week.

ascjones: In the end I went for repeating the subscription name, since it was easier to implement with what I had already done.

ascjones:
Now ready for full review @tomusdrw

derive/tests/ui/pubsub/missing-subscribe.rs
```diff
@@ -0,0 +1,18 @@
+#[macro_use]
+extern crate jsonrpc_derive;
+extern crate jsonrpc_core;
+extern crate jsonrpc_pubsub;
+
+#[rpc]
+pub trait Rpc {
+	type Metadata;
+
+//	#[pubsub(subscription = "hello", name = "hello_subscribe", alias("hello_sub"))]
```
tomusdrw: let's remove it maybe? Or add a comment that this is missing

derive/tests/ui/pubsub/missing-unsubscribe.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[rpc]
+pub trait Rpc {
+	type Metadata;
+
+	/// Hello subscription
+	#[pubsub(subscription = "hello", subscribe, name = "hello_subscribe", alias("hello_sub"))]
+	fn subscribe(&self, Self::Metadata, typed::Subscriber<String>, u64);
+
+//	#[pubsub(subscription = "hello", unsubscribe, name = "hello_unsubscribe")]
```
tomusdrw: same as above

pubsub/src/typed.rs
```diff
@@ -0,0 +1,135 @@
+//! PUB-SUB auto-serializing structures.
+
+use std::marker::PhantomData;
+
+use serde;
+//use util::Fn;
```
tomusdrw: uncomment / remove

PR #350, opened 2019-01-03
case insensitive charset matching

### LINKED ISSUE #349: charset value handing does not follow http spec
in https://tools.ietf.org/html/rfc2616#section-3.4, 
> HTTP character sets are identified by case-insensitive tokens.

but in handler.rs, 
> 		match content_type.and_then(|val| val.to_str().ok()) {
>			Some("application/json") => true,
>			Some("application/json; charset=utf-8") => true,
only "utf-8" is considered as valid, and only one space is allowed. This causes lots of request failed.

seunlanlege:
closes #349

### DIFF
--- a/http/src/handler.rs
+++ b/http/src/handler.rs
@@ -565,10 +565,11 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
 		match content_type.and_then(|val| val.to_str().map(str::to_lowercase).ok()) {
 			Some(ref content)
 				if content.eq_ignore_ascii_case("application/json")
-					|| content.replace(" " ,"").eq_ignore_ascii_case("application/json;charset=utf-8") =>
+					|| content.eq_ignore_ascii_case("application/json; charset=utf-8")
+					|| content.eq_ignore_ascii_case("application/json;charset=utf-8") =>
 			{
 				true
-			}
+			},
 			_ => false,
 		}
 	}
@@ -586,6 +587,11 @@ mod test {
 			.body(())
 			.unwrap();
 
+		let request2 = hyper::Request::builder()
+			.header("content-type", "Application/Json;charset=UTF-8")
+			.body(())
+			.unwrap();
+
 
 		assert_eq!(
 			request.headers().get("content-type").unwrap(),
@@ -594,6 +600,10 @@ mod test {
 
 		assert_eq!(
 			RpcHandler::<(), Noop>::is_json(request.headers().get("content-type")),
+			true
+		);
+		assert_eq!(
+			RpcHandler::<(), Noop>::is_json(request2.headers().get("content-type")),
 			true
 		);
 	}
--- a/http/src/handler.rs
+++ b/http/src/handler.rs
@@ -1,18 +1,18 @@
 use Rpc;
 
-use std::{fmt, mem, str};
 use std::sync::Arc;
+use std::{fmt, mem, str};
 
-use hyper::{self, service::Service, Body, Method};
 use hyper::header::{self, HeaderMap, HeaderValue};
+use hyper::{self, service::Service, Body, Method};
 
-use jsonrpc::{self as core, middleware, FutureResult, Metadata, Middleware};
-use jsonrpc::futures::{Future, Poll, Async, Stream, future};
+use jsonrpc::futures::{future, Async, Future, Poll, Stream};
 use jsonrpc::serde_json;
+use jsonrpc::{self as core, middleware, FutureResult, Metadata, Middleware};
 use response::Response;
 use server_utils::cors;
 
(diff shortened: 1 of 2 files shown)


http/src/handler.rs
```diff
@@ -562,10 +616,58 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
...
+	use jsonrpc_core::*;
+	use std::sync::atomic::{AtomicUsize};
+	
+	#[derive(Clone, Debug)]
+	struct Meta(usize);
+	
+	impl Metadata for Meta {}
+
+	#[derive(Default)]
+	struct MyMiddleware(AtomicUsize);
```
tomusdrw: Use `middleware::Noop`

http/src/handler.rs
```diff
@@ -562,10 +616,58 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
...
 		}
 	}
 }
+
+#[cfg(test)]
+mod test {
+	use super::{hyper, RpcHandler};
+	use jsonrpc_core::futures::future::Either;
+	use jsonrpc_core::futures::Future;
+	use jsonrpc_core::*;
```
tomusdrw: Please avoid glob imports.

http/src/handler.rs
```diff
@@ -562,10 +616,58 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
...
+			X: Future<Item = Option<Response>, Error = ()> + Send + 'static,
+		{
+			Either::A(Box::new(next(request, meta)))
+		}
+	}
+
+
+	#[test]
+	fn test_case_insensitive_content_type() {
+		let request = hyper::Request::builder()
```
tomusdrw: Can we please add another assertion to make sure that the builder doesn't normalize the value?

http/src/handler.rs
```diff
@@ -562,10 +562,45 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
 	/// Returns true if the `content_type` header indicates a valid JSON
 	/// message.
 	fn is_json(content_type: Option<&header::HeaderValue>) -> bool {
-		match content_type.and_then(|val| val.to_str().ok()) {
-			Some("application/json") => true,
-			Some("application/json; charset=utf-8") => true,
+		match content_type.and_then(|val| val.to_str().map(str::to_lowercase).ok()) {
+			Some(ref content)
```
tomusdrw: Just noticed that we could avoid allocation of `to_lowercase` by just using `str::eq_ignore_ascii_case`.

http/src/handler.rs
```diff
@@ -562,10 +562,45 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
...
+				if content == "application/json"
+					|| content == "application/json; charset=utf-8" =>
```
tomusdrw: Is the space relevant here? Should we match against no-space variant as well?

http/src/handler.rs
```diff
@@ -562,10 +562,45 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
...
 }
+
+#[cfg(test)]
+mod test {
+	use super::{hyper, RpcHandler};
+	use jsonrpc_core::middleware::Noop;
+	use jsonrpc_core::Metadata;
+	
+	#[derive(Clone, Debug)]
+	struct Meta(usize);
```
tomusdrw: You can also use `()` instead of `Meta` instead of this dummy struct

http/src/handler.rs
seunlanlege: the spec doesn't mention spaces, but to be on the safe side, let's match without spaces

http/src/handler.rs
```diff
@@ -562,10 +562,39 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
 	/// Returns true if the `content_type` header indicates a valid JSON
 	/// message.
 	fn is_json(content_type: Option<&header::HeaderValue>) -> bool {
-		match content_type.and_then(|val| val.to_str().ok()) {
-			Some("application/json") => true,
-			Some("application/json; charset=utf-8") => true,
+		match content_type.and_then(|val| val.to_str().map(str::to_lowercase).ok()) {
+			Some(ref content)
+				if content.eq_ignore_ascii_case("application/json")
+					|| content.replace(" " ,"").eq_ignore_ascii_case("application/json;charset=utf-8") =>
```
tomusdrw: `replace` allocates a new string, let's duplicate the check maybe?

leoaion:
Thanks for the quick fix. will you merge this fix to branch parity-2.2?

leoaion:
will this fix be merged into parity-2.2 branch? or under any other released branch/tag?

tomusdrw:
@leoaion yes, see backports here #350 later on it will get released as well.

PR #351, opened 2019-01-04
Backports to Parity-2.2

tomusdrw:


http/src/handler.rs
```diff
@@ -562,10 +562,49 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
 	/// Returns true if the `content_type` header indicates a valid JSON
 	/// message.
 	fn is_json(content_type: Option<&header::HeaderValue>) -> bool {
-		match content_type.and_then(|val| val.to_str().ok()) {
-			Some("application/json") => true,
-			Some("application/json; charset=utf-8") => true,
+		match content_type.and_then(|val| val.to_str().map(str::to_lowercase).ok()) {
```
seunlanlege:
```suggestion
		match content_type.and_then(|val| val.to_str().ok()) {
```

tomusdrw: @seunlanlege this is just a backports PR, can you submit a PR to fix it on master so that we can later backport it?

seunlanlege: alright

PR #356, opened 2019-01-08
Solve more issues with commas.

tomusdrw:
Follow up on #355 there was even more commas misplaced.
Now we don't require trailing commas only in the initial parsing (before adding `@*` tokens) - after that the commas are required after every element.

I've added a test case with Metadata as well, cause it's parsed separately.


tomusdrw:
Windows failure unrelated.

PR #359, opened 2019-01-12
Derive Clone for all core types::*

passcod:
I needed Clone on `MethodCall` for an implementation, and I can't see a reason for not having it on the other types as well.


tomusdrw:
thanks!

PR #365, opened 2019-01-29
Support multiple trailing arguments

svyatonik:
The rules are exactly as in C++: if you need to pass value of some optional argument, you must also pass values for all previous optional arguments. So there's no ambiguities with that.

This PR supersedes the `pzec_dependency_multiple_trailing_args` branch, but please do not delete it - we'll need to update parity-zcash once/if this PR is merged. I'll delete it by myself later.


svyatonik:
cc @ascjones

PR #371, opened 2019-01-30
One way serialize for subscriber type

### LINKED ISSUE #370: Deserialize trait shouldn't be required for T in Subscriber<T>
`Deserialize` should only be required for types which are passed as params to rpc methods. A pubsub `subscribe` method accepts a `Subscriber<T>` and the visitor does not discriminate so assumes it requires the type bound.

Found when converting `substrate` to use `jsonrpc-derive`

```
Compiling substrate-rpc-servers v0.1.0 (/home/andrew/code/paritytech/substrate/core/rpc-servers)
error[E0277]: the trait bound `for<'de> <Block as sr_primitives::traits::Block>::Header: serde::Deserialize<'de>` is not satisfied
  --> core/rpc-servers/src/lib.rs:59:23
   |
59 |     io.extend_with(chain.to_delegate());
   |                          ^^^^^^^^^^^ the trait `for<'de> serde::Deserialize<'de>` is not implemented for `<Block as sr_primitives::traits::Block>::Header`
   |
   = help: consider adding a `where for<'de> <Block as sr_primitives::traits::Block>::Header: serde::Deserialize<'de>` bound
   = note: required because of the requirements on the impl of `sr_primitives::DeserializeOwned` for `<Block as sr_primitives::traits::Block>::Header`
```

Fix will be somewhere in the visitor here: https://github.com/paritytech/jsonrpc/blob/master/derive/src/to_delegate.rs#L360

ascjones:
Fixes #370

### DIFF
--- a/derive/tests/run-pass/pubsub-subscription-type-without-deserialize.rs
+++ b/derive/tests/run-pass/pubsub-subscription-type-without-deserialize.rs
@@ -1,0 +1,29 @@
+#[macro_use]
+extern crate serde_derive;
+
+use jsonrpc_core::Result;
+use jsonrpc_derive::rpc;
+use jsonrpc_pubsub::{typed::Subscriber, SubscriptionId};
+
+// One way serialization
+#[derive(Serialize)]
+struct SerializeOnly {
+	foo: String,
+}
+
+#[rpc]
+pub trait Rpc {
+	type Metadata;
+
+	/// Hello subscription
+	#[pubsub(subscription = "hello", subscribe, name = "hello_subscribe", alias("hello_sub"))]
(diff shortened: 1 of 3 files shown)


derive/tests/run-pass/pubsub-subscription-type-without-deserialize.rs
```diff
@@ -0,0 +1,49 @@
+extern crate serde;
+#[macro_use]
+extern crate serde_derive;
+extern crate jsonrpc_core;
+extern crate jsonrpc_pubsub;
+#[macro_use]
+extern crate jsonrpc_derive;
+
```
ascjones: I tried to get `--edition=2018` working with compiletests, but couldn't manage

PR #375, opened 2019-02-05
Fix rpc alias(es) key constant

ascjones:
Fixes https://github.com/paritytech/substrate/issues/1683. /cc @bkchr

### DIFF
--- a/derive/tests/macros.rs
+++ b/derive/tests/macros.rs
@@ -22,7 +22,7 @@ pub trait Rpc {
 	fn neg(&self, _: i64) -> Result<i64>;
 
 	/// Adds two numbers and returns a result
-	#[rpc(name = "add")]
+	#[rpc(name = "add", alias("add_alias1", "add_alias2"))]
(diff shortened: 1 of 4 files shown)


derive/tests/macros.rs
```diff
@@ -22,7 +22,7 @@ pub trait Rpc {
 	fn neg(&self, _: i64) -> Result<i64>;
 
 	/// Adds two numbers and returns a result
-	#[rpc(name = "add")]
+	#[rpc(name = "add", alias("add_alias1", "add_alias2"))]
```
tomusdrw: Should we fail the derivation in case there are unknown parameters in here? Like afaiu we had `alias` which was something that we don't expect (cause we expected `aliases`), so IMHO this should just error during compilation.

ascjones: Agreed: see #376. I could add it in to this PR if you like? Was just getting this fix in quickly for substrate#1683

tomusdrw: Nope, it's fine. Haven't noticed it's already logged.

Issue #380, opened 2019-02-07
derive: generate nice compiler error messages using `syn::Error`


ascjones:
Currently we just panic at the top level when there is an `Error`: https://github.com/paritytech/jsonrpc/blob/master/derive/src/lib.rs#L149.

We should rather use (`syn::Error`) and then `e.to_compile_error()` - e.g. https://github.com/paritytech/substrate/blob/master/core/sr-api-macros/src/impl_runtime_apis.rs#L614

We should be able to pass the relevant `span` when instantiating the error, which should produce a nice error message.

Hat tip @bkchr.

PR #382, opened 2019-02-07
use compile error

thiolliere:
related to https://github.com/paritytech/jsonrpc/issues/380


ascjones:
That was quick - thanks @thiolliere

Issue #376, opened 2019-02-05
derive: should fail to compile with invalid attribute args


ascjones:
There was a bug where I had changed the key constant so aliases are being ignored: https://github.com/paritytech/substrate/issues/1683. Fixed in https://github.com/paritytech/jsonrpc/pull/375.

However if there is an invalid arg to the attribute it should fail to compile in the first place, not just ignore it as it currently does, so we can catch such issues at compile time.

PR #384, opened 2019-02-08
derive: don't require pubsub dependency if no pubsub methods

### LINKED ISSUE #369: Don't require pubsub dependency when using derive if pubsub not used
If you use `#[rpc]` from `derive` but don't have any `pubsub` methods then you shouldn't need a dependency on `pubsub`.

Would need to remove this import if no pubsub methods: https://github.com/paritytech/jsonrpc/blob/master/derive/src/rpc_trait.rs#L161

rel #112

ascjones:
Fixes #369

### DIFF
--- a/derive/src/rpc_trait.rs
+++ b/derive/src/rpc_trait.rs
@@ -53,7 +53,7 @@ impl<'a> Fold for RpcTrait {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ascjones:
Yay semver. If we're doing a release I'll quickly get #376 in, then we can probably justify a `10.1`

PR #385, opened 2019-02-08
Fail to compile with invalid attribute parameters

### LINKED ISSUE #376: derive: should fail to compile with invalid attribute args

ascjones:
Closes #376. rel: #380.

If there are any invalid attribute params it will fail to compile, rather than just ignore as it did before. This would have caught issues like https://github.com/paritytech/substrate/issues/1683 at compile time.


derive/src/rpc_attr.rs
```diff
@@ -113,6 +120,58 @@ impl RpcMethodAttribute {
...
+		_ => Ok(meta), // ignore other attributes - compiler will catch unknown ones
+	}
+}
+
+fn validate_idents(meta: &syn::Meta, attr_idents: &[String], valid: &[&str]) -> Result<syn::Meta> {
+	let invalid_meta_words: Vec<_> = attr_idents
+		.iter()
+		.cloned()
+		.filter(|w| !valid.iter().any(|v| v == w))
+		.collect();
```
niklasad1: Isn't better to `filter()` before `clone()` to avoid less allocation?

However, I haven't checked to assembly but I suspect `rustc` is smart enough to re-arrange those

derive/src/rpc_attr.rs
```diff
@@ -27,24 +29,29 @@ const METADATA_META_WORD: &'static str = "meta";
 const SUBSCRIBE_META_WORD: &'static str = "subscribe";
 const UNSUBSCRIBE_META_WORD: &'static str = "unsubscribe";
 
+const MULTIPLE_RPC_ATTRIBUTES_ERR: &'static str = "Expected only a single rpc attribute per method";
+const INVALID_ATTR_PARAM_NAMES_ERR: &'static str = "Invalid attribute parameter(s):";
+const MISSING_NAME_ERR: &'static str = "rpc attribute should have a name e.g. `name = \"method_name\"`";
+const MISSING_SUB_NAME_ERR: &'static str = "pubsub attribute should have a subscription name";
+const BOTH_SUB_AND_UNSUB_ERR: &'static str = "pubsub attribute annotated with both subscribe and unsubscribe";
+const NEITHER_SUB_OR_UNSUB_ERR: &'static str = "pubsub attribute not annotated with either subscribe or unsubscribe";
+
```
niklasad1: it is enough with `&str` here

PR #399, opened 2019-02-28
ws: replace error-chain with vanilla Error impl

ascjones:
Just did this quickly to see what it is like converting an `error-chain` to vanilla Error impl. (because of `Error::cause` being deprecated in Rust 1.33)

This is one option to consider for https://github.com/paritytech/parity-ethereum/issues/10302 and https://github.com/paritytech/substrate/issues/1547

### DIFF
--- a/ws/src/error.rs
+++ b/ws/src/error.rs
@@ -1,28 +1,49 @@
 #![allow(missing_docs)]
 
-use std::io;
+use std::{io, error, fmt, result};
 
 use crate::ws;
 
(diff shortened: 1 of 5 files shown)


ws/src/error.rs
```diff
@@ -1,28 +1,49 @@
 #![allow(missing_docs)]
```
tomusdrw: I guess we don't need this now and can add the missing docs.

PR #401, opened 2019-03-08
Add asynchronous methods to assign_id or reject in pubsub.

tomusdrw:
Related https://github.com/polkadot-js/api/issues/764

This will allow us to only start sending notifications after we are (semi) sure that the id has actually been already sent.

I kept the old methods to avoid breaking the API as the new one is slightly more tricky to use, while imho is not super critical in most scenarios.

Also note that we can't just use `.wait()` on the async API, cause:
1. It's not recommended anyway since it blocks the event loop
2. Will always block indefinitely if run on the same frame because the response can never be sent since the generation part is blocking it.


pubsub/src/typed.rs
```diff
@@ -40,17 +40,37 @@ impl<T, E> Subscriber<T, E> {
 		self.subscriber.reject(error)
 	}
 
+	/// Reject subscription with given error.
+	///
+	/// The returned futuer will resolve when the response is sent to the client.
```
ascjones:
```suggestion
	/// The returned future will resolve when the response is sent to the client.
```

PR #405, opened 2019-03-22
Implement client support

### LINKED ISSUE #253: Implement Error for core::Error
Hello,

while migrating an application from `error-chain` to `failure` I noticed that `core::Error` doesn't implement `Error`.
I can't define return type as `Result<T, failure::Error>` because of that.

I would like to know are there any problems which block us from this implementation?
If this could be done I would like to try to work on it.

Thanks.

P.S.
I also asked on [forum](https://users.rust-lang.org/t/migrate-application-from-error-chain-to-failure/16601) if I work with errors right.
Maybe I just miss something.
Also I am not sure if implementing `Error` would be useful at all.

### LINKED ISSUE #381: Create client side interface
With the new derive crate it should be simple to provide a client interface as well. The client side would look like:
```
trait Client {
    fn my_function(&self, data: u32) -> Future<String>;
}
```

Calling `my_function` will call the RPC method on the server.

dvc94ch:
Requirements for generating a client are:
- rpc trait methods need to name their arguments
- pubsub methods are ignored

closes #253 and #381

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Parity Technologies <admin@parity.io>"]
 description = "Transport agnostic JSON-RPC 2.0 client implementation."
-documentation = "https://docs.rs/jsonrpc-core/"
+documentation = "https://docs.rs/jsonrpc-client/"
 edition = "2018"
 homepage = "https://github.com/paritytech/jsonrpc"
 keywords = ["jsonrpc", "json-rpc", "json", "rpc", "serde"]
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -27,6 +27,7 @@ serde_json = "1.0"
 
 [dev-dependencies]
 jsonrpc-derive = { version = "10.1", path = "../derive" }
+jsonrpc-client = { version = "10.1", path = "." }
 serde = "1.0"
 tokio = "0.1"
 
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -1,0 +1,34 @@
+[package]
+authors = ["Parity Technologies <admin@parity.io>"]
+description = "Transport agnostic JSON-RPC 2.0 client implementation."
+documentation = "https://docs.rs/jsonrpc-core/"
+edition = "2018"
+homepage = "https://github.com/paritytech/jsonrpc"
+keywords = ["jsonrpc", "json-rpc", "json", "rpc", "serde"]
+license = "MIT"
+name = "jsonrpc-client"
+repository = "https://github.com/paritytech/jsonrpc"
+version = "10.1.0"
+
+categories = [
+	"asynchronous",
+	"network-programming",
+	"web-programming::http-client",
+	"web-programming::http-server",
+	"web-programming::websocket",
+]
+
+[dependencies]
+failure = "0.1"
+futures = "~0.1.6"
+jsonrpc-core = { version = "10.1", path = "../core" }
+log = "0.4"
+serde_json = "1.0"
+
+[dev-dependencies]
+jsonrpc-derive = { version = "10.1", path = "../derive" }
+serde = "1.0"
+tokio = "0.1"
+
+[badges]
+travis-ci = { repository = "paritytech/jsonrpc", branch = "master"}
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -58,7 +58,7 @@ impl Future for RpcFuture {
 	type Error = RpcError;
 
 	fn poll(&mut self) -> Result<Async<Self::Item>, Self::Error> {
-		// TODO should timeout
+		// TODO should timeout (#410)
 		match self.recv.poll() {
 			Ok(Async::Ready(Ok(value))) => Ok(Async::Ready(value)),
 			Ok(Async::Ready(Err(error))) => Err(RpcError::JsonRpcError(error)),
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -1,0 +1,328 @@
+//! JSON-RPC client implementation.
+#![deny(missing_docs)]
+
+use jsonrpc_core::{
+	Call, Error, Id, MethodCall, Output, Params,
+	Request, Response, Version,
+};
+use futures::prelude::*;
+use futures::sync::{mpsc, oneshot};
+use failure::{Fail, format_err};
+use log::debug;
+use serde_json::Value;
+use std::collections::HashMap;
+use std::collections::VecDeque;
+
+/// The errors returned by the client.
+#[derive(Debug, Fail)]
+pub enum RpcError {
+	/// An error returned by the server.
+	#[fail(display = "Server returned rpc error {}", _0)]
+	JsonRpcError(Error),
+	/// Failure to parse server response.
+	#[fail(display = "Failed to parse server response")]
+	ParseError,
+	/// Request timed out.
+	#[fail(display = "Request timed out")]
+	Timeout,
+	/// The server returned a response with an unknown id.
+	#[fail(display = "Server returned a response with an unknown id")]
+	UnknownId,
+	/// Not rpc specific errors.
+	#[fail(display = "{}", _0)]
+	Other(failure::Error),
+}
+
+impl From<Error> for RpcError {
+	fn from(error: Error) -> Self {
+		RpcError::JsonRpcError(error)
+	}
+}
+
+/// The future retured by the client.
+pub struct RpcFuture {
+	recv: oneshot::Receiver<Result<Value, Error>>,
+}
+
+impl RpcFuture {
+	/// Creates a new `RpcFuture`.
+	pub fn new(recv: oneshot::Receiver<Result<Value, Error>>) -> Self {
+		RpcFuture {
+			recv
+		}
+	}
+}
+
+impl Future for RpcFuture {
+	type Item = Value;
+	type Error = RpcError;
+
+	fn poll(&mut self) -> Result<Async<Self::Item>, Self::Error> {
+		// TODO should timeout
+		match self.recv.poll() {
+			Ok(Async::Ready(Ok(value))) => Ok(Async::Ready(value)),
+			Ok(Async::Ready(Err(error))) => Err(RpcError::JsonRpcError(error)),
+			Ok(Async::NotReady) => Ok(Async::NotReady),
+			Err(error) => Err(RpcError::Other(error.into())),
+		}
+	}
+}
+
+/// A message sent to the `RpcClient`. This is public so that
+/// the derive crate can generate a client.
+pub struct RpcMessage {
+	/// The rpc method name.
+	pub method: String,
+	/// The rpc method parameters.
+	pub params: Params,
+	/// The oneshot channel to send the result of the rpc
+	/// call to.
+	pub sender: oneshot::Sender<Result<Value, Error>>,
+}
+
+/// A channel to a `RpcClient`.
+pub type RpcChannel = mpsc::Sender<RpcMessage>;
+
+/// The RpcClient handles sending and receiving asynchronous
+/// messages through an underlying transport.
+pub struct RpcClient<TSink, TStream> {
+	id: u64,
+	queue: HashMap<Id, oneshot::Sender<Result<Value, Error>>>,
+	sink: TSink,
+	stream: TStream,
+	channel: Option<mpsc::Receiver<RpcMessage>>,
+	outgoing: VecDeque<String>,
+}
+
+impl<TSink, TStream> RpcClient<TSink, TStream> {
+	/// Creates a new `RpcClient`.
+	pub fn new(
+		sink: TSink,
+		stream: TStream,
+		channel: mpsc::Receiver<RpcMessage>,
+	) -> Self {
+		RpcClient {
+			id: 0,
+			queue: HashMap::new(),
+			sink,
+			stream,
+			channel: Some(channel),
+			outgoing: VecDeque::new(),
+		}
+	}
+
+	fn next_id(&mut self) -> Id {
+		let id = self.id;
+		self.id = id + 1;
+		Id::Num(id)
+	}
+}
+
+impl<TSink, TStream> Future for RpcClient<TSink, TStream>
+where
+	TSink: Sink<SinkItem=String, SinkError=RpcError>,
+	TStream: Stream<Item=String, Error=RpcError>,
+{
+	type Item = ();
+	type Error = RpcError;
+
+	fn poll(&mut self) -> Result<Async<Self::Item>, Self::Error> {
+		// Handle requests from the client.
+		loop {
+			if self.channel.is_none() {
+				break;
+			}
+			let msg = match self.channel.as_mut().unwrap().poll() {
+				Ok(Async::Ready(Some(msg))) => msg,
+				Ok(Async::Ready(None)) => {
+					// When the channel is dropped we still need to finish
+					// outstanding requests.
+					self.channel.take();
+					break;
+				},
+				Ok(Async::NotReady) => break,
+				Err(()) => continue,
+			};
+			let id = self.next_id();
+			let request = Request::Single(Call::MethodCall(MethodCall {
+				jsonrpc: Some(Version::V2),
+				method: msg.method,
+				params: msg.params,
+				id: id.clone(),
+			}));
+			self.queue.insert(id, msg.sender);
+			let request_str = serde_json::to_string(&request)
+				.map_err(|error| RpcError::Other(error.into()))?;
+			self.outgoing.push_back(request_str);
+		}
+		// Handle outgoing rpc requests.
+		loop {
+			match self.outgoing.pop_front() {
+				Some(request) => {
+					match self.sink.start_send(request)? {
+						AsyncSink::Ready => {},
+						AsyncSink::NotReady(request) => {
+							self.outgoing.push_front(request);
+							break;
+						},
+					}
+				}
+				None => break,
+			}
+		}
+		let done_sending = match self.sink.poll_complete()? {
+			Async::Ready(()) => true,
+			Async::NotReady => false,
+		};
+		// Handle incoming rpc requests.
+		loop {
+			let response_str = match self.stream.poll() {
+				Ok(Async::Ready(Some(response_str))) => response_str,
+				Ok(Async::Ready(None)) => {
+					// The websocket connection was closed so the client
+					// can be shutdown. Reopening closed connections must
+					// be handled by the transport.
+					debug!("connection closed");
+					return Ok(Async::Ready(()))
+				},
+				Ok(Async::NotReady) => break,
+				Err(err) => Err(err)?,
+			};
+			let response = serde_json::from_str::<Response>(&response_str)
+				.map_err(|error| RpcError::Other(error.into()))?;
+			let outputs: Vec<Output> = match response {
+				Response::Single(output) => vec![output],
+				Response::Batch(outputs) => outputs,
+			};
+			for output in outputs {
+				let channel = self.queue.remove(output.id());
+				let value: Result<Value, Error> = output.into();
+				match channel {
+					Some(tx) => tx.send(value)
+						.map_err(|_| {
+							RpcError::Other(format_err!("oneshot channel closed"))
+						})?,
+					None => Err(RpcError::UnknownId)?,
+				};
+			}
+		}
+		if self.channel.is_none() && self.outgoing.is_empty() && self.queue.is_empty() && done_sending {
+			debug!("client finished");
+			Ok(Async::Ready(()))
+		} else {
+			Ok(Async::NotReady)
+		}
+	}
+}
+
+/// Utilities for testing.
+pub mod test {
+	use super::*;
+	use jsonrpc_core::{IoHandler, RemoteProcedure};
+
+	/// Implements a rpc client using an `IoHandler`.
+	pub struct FakeRpc {
+		handler: IoHandler,
+		queue: VecDeque<String>,
+	}
+
+	impl FakeRpc {
+		/// Creates a new `FakeRpc`.
+		pub fn new(handler: IoHandler) -> Self {
+			FakeRpc {
+				handler,
+				queue: VecDeque::new(),
+			}
+		}
+	}
+
+	impl Stream for FakeRpc {
+		type Item = String;
+		type Error = RpcError;
+
+		fn poll(&mut self) -> Result<Async<Option<Self::Item>>, Self::Error> {
+			match self.queue.pop_front() {
+				Some(response) => Ok(Async::Ready(Some(response))),
+				None => Ok(Async::NotReady),
+			}
+		}
+	}
+
+	impl Sink for FakeRpc {
+		type SinkItem = String;
+		type SinkError = RpcError;
+
+		fn start_send(
+			&mut self,
+			request: Self::SinkItem,
+		) -> Result<AsyncSink<Self::SinkItem>, Self::SinkError> {
+			match self.handler.handle_request_sync(&request) {
+				Some(response) => self.queue.push_back(response),
+				None => {},
+			};
+			Ok(AsyncSink::Ready)
+		}
+
+		fn poll_complete(&mut self) -> Result<Async<()>, Self::SinkError> {
+			Ok(Async::Ready(()))
+		}
+	}
+
+	/// Connects to a `IoHandler`.
+	pub fn connect<TClient, TServer>(
+		server: TServer,
+	) -> (TClient, impl Future<Item=(), Error=RpcError>)
+	where
+		TClient: From<RpcChannel>,
+		TServer: Into<HashMap<String, RemoteProcedure<()>>>,
+	{
+		let mut io = IoHandler::new();
+		io.extend_with(server);
+		let (sink, stream) = test::FakeRpc::new(io).split();
+		let (sender, receiver) = mpsc::channel(0);
+		let rpc_client = RpcClient::new(sink, stream, receiver);
+		let client = TClient::from(sender);
+		(client, rpc_client)
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use crate as jsonrpc_client;
+	use super::*;
+	use jsonrpc_core::Result;
+	use jsonrpc_derive::rpc;
+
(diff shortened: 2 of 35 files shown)


dvc94ch:
[clabot:check]

derive/examples/std.rs
```diff
@@ -5,32 +5,32 @@ use jsonrpc_derive::rpc;
...
+	fn call(&self, a: u64) -> FutureResult<String, Error>;
 }
 
 struct RpcImpl;
 
 impl Rpc for RpcImpl {
-	fn protocol_version(&self) -> Result<String> {
-		Ok("version1".into())
-	}
+    fn protocol_version(&self) -> Result<String> {
```
tomusdrw: We use tabs in the entire codebase, could you reformat all files?

derive/src/rpc_trait.rs
```diff
@@ -131,13 +132,15 @@ fn generate_rpc_item_trait(item_trait: &syn::ItemTrait) -> Result<(syn::ItemTrai
 
 	let to_delegate_method =
 		generate_trait_item_method(&method_registrations, &item_trait, rpc_trait.has_metadata, has_pubsub_methods)?;
+
+  let rpc_client_module = generate_client_module(&method_registrations, &item_trait)?;
```
tomusdrw: Can we make this optional, maybe a parameter to the main `#[rpc]` attribute? Otherwise it will generate unused code warnings, no?

derive/src/to_client.rs
```diff
@@ -0,0 +1,134 @@
...
+                    }).collect();
+                method_decl.push(syn::parse_quote! {
+                    fn #name(#args) -> #new_ty;
+                });
+                method_impl.push(syn::parse_quote! {
+                    fn #name(#args) -> #new_ty {
+                        let args_tuple = (#(#arg_names,)*);
+                        let args = serde_json::to_value(args_tuple)
+                            .expect("should never fail");
+                        let params = serde_json::from_value(args)
```
tomusdrw: Can't we go directly to `Params`? We always expect this to be an array, no?

derive/src/to_client.rs
```diff
@@ -0,0 +1,134 @@
...
+            _ => {}
+        }
+    }
+
+    let mut rpc_client_trait = rpc_trait.to_owned();
+    rpc_client_trait.items = method_decl;
+
+    let trait_name = &rpc_client_trait.ident;
+    let rpc_client_impl: syn::ItemImpl = syn::parse_quote! {
+        impl<T: jsonrpc_core::client::RpcClient> #trait_name for T {
```
tomusdrw: Do we really need a trait for this? While I think it's nice for the server side, when the actual implementation has to be provided by the user, I don't think it's useful in case we generate the implementation anyway.

Maybe let's just stick to regular struct + methods?

derive/src/to_client.rs
```diff
@@ -0,0 +1,134 @@
...
+            MethodRegistration::Standard { method, .. } => {
+                let rpc_name = method.name();
+                let name = &method.trait_item.sig.ident;
+                let returns = match &method.attr.kind {
+                    AttributeKind::Rpc { returns: Some(returns), .. } => returns,
+                    _ => continue,
+                };
+                let returns: syn::Type = syn::parse_str(returns)?;
+                let new_ty: syn::Type = syn::parse_quote! {
+                    future::AndThen<
```
tomusdrw: We can just return `impl Future` in case we ditch the client trait.

derive/src/to_client.rs
```diff
@@ -0,0 +1,134 @@
...
+                            Response::Single(output) => output.into(),
+                            _ => return future::err(Error::parse_error()),
+                        };
+                        let value = match value {
+                            Ok(value) => value,
+                            Err(error) => return future::err(error),
+                        };
+                        let result = serde_json::from_value::<#returns>(value)
+                            .map_err(|_| Error::new(ErrorCode::ParseError));
+                        match result {
```
tomusdrw: `future::done(result)` ?

derive/src/to_client.rs
```diff
@@ -0,0 +1,134 @@
...
+                        let value: Result<Value> = match response {
+                            Response::Single(output) => output.into(),
+                            _ => return future::err(Error::parse_error()),
+                        };
+                        let value = match value {
+                            Ok(value) => value,
+                            Err(error) => return future::err(error),
+                        };
+                        let result = serde_json::from_value::<#returns>(value)
+                            .map_err(|_| Error::new(ErrorCode::ParseError));
```
tomusdrw: Might be useful to actually log the original error here

derive/src/to_client.rs
```diff
@@ -0,0 +1,134 @@
...
+                        let request_str = serde_json::to_string(&request)
+                            .expect("should never fail");
+                        self.call_method(request_str)
+                            .and_then(rpc_impl_response::#name)
+                    }
+                });
+                method_resp.push(syn::parse_quote! {
+                    pub fn #name(response_str: String) -> FutureResult<#returns, Error> {
+                        let response = serde_json::from_str::<Response>(&response_str)
+                            .map_err(|_| Error::new(ErrorCode::ParseError));
```
tomusdrw: Might be useful to actually log the original error here

derive/src/to_client.rs
```diff
@@ -0,0 +1,134 @@
...
+                        let response = match response {
+                            Ok(response) => response,
+                            Err(error) => return future::err(error),
+                        };
+                        let value: Result<Value> = match response {
+                            Response::Single(output) => output.into(),
+                            _ => return future::err(Error::parse_error()),
```
tomusdrw: Maybe produce a warning to make it easier to debug potential issues

derive/src/to_client.rs
tomusdrw: Might be useful to actually log the original error here

derive/src/to_client.rs
```diff
@@ -0,0 +1,134 @@
...
+                        let args_tuple = (#(#arg_names,)*);
+                        let args = serde_json::to_value(args_tuple)
+                            .expect("should never fail");
+                        let params = serde_json::from_value(args)
+                            .expect("should never fail");
+                        let request = Request::Single(Call::MethodCall(MethodCall {
+                            jsonrpc: Some(Version::V2),
+                            method: #rpc_name.to_owned(),
+                            params,
+                            id: Id::Num(next_id()),
```
tomusdrw: Seems that we don't really use `id` anyway, do we?

core/src/client.rs
```diff
@@ -0,0 +1,16 @@
...
+/// The RpcClient trait. Use the #[rpc] macro derives the Request and Response
+/// generation.
+pub trait RpcClient: Send + Sync + 'static {
+    /// Calls an rpc method.
+    fn call_method(&self, request: String) -> FutureResult<String, Error>;
+}
+
+impl RpcClient for IoHandler {
+    fn call_method(&self, request: String) -> FutureResult<String, Error> {
+        future::ok(self.handle_request_sync(&request).unwrap())
```
tomusdrw: We can't be sure that `request` is not a `Notification` (no id), so the `unwrap` here seems invalid. Perhaps turn it into an error.

derive/examples/std.rs
```diff
@@ -5,32 +5,32 @@ use jsonrpc_derive::rpc;
 #[rpc]
 pub trait Rpc {
 	/// Returns a protocol version
-	#[rpc(name = "protocolVersion")]
+	#[rpc(name = "protocolVersion", returns = "String")]
```
tomusdrw: While I understand that it's not possible to auto-derive the return types, I'd like to propose two things:
1. Require `returns` attribute only when client is being generated (see my other suggestion to add a parameter to `#[rpc]`)
2. Deduce the return type for types like `Result`, `Future`, etc and allow overriding using `returns`. This will greatly simplify most common cases. I'm fine to have that as a separate PR or just as an open issue.

derive/src/to_client.rs
```diff
@@ -0,0 +1,134 @@
...
+                        }
+                    }).collect();
+                method_decl.push(syn::parse_quote! {
+                    fn #name(#args) -> #new_ty;
+                });
+                method_impl.push(syn::parse_quote! {
+                    fn #name(#args) -> #new_ty {
+                        let args_tuple = (#(#arg_names,)*);
+                        let args = serde_json::to_value(args_tuple)
+                            .expect("should never fail");
```
tomusdrw:
```suggestion
                            .expect("Only types with infallible serialisation can be used for JSON-RPC");
```

derive/src/to_client.rs
```diff
@@ -0,0 +1,134 @@
...
+                        let params = serde_json::from_value(args)
+                            .expect("should never fail");
+                        let request = Request::Single(Call::MethodCall(MethodCall {
+                            jsonrpc: Some(Version::V2),
+                            method: #rpc_name.to_owned(),
+                            params,
+                            id: Id::Num(next_id()),
+                        }));
+                        let request_str = serde_json::to_string(&request)
+                            .expect("should never fail");
```
tomusdrw:
```suggestion
                            .expect("internal JSON-RPC types have infallible serialisation; qed");
```

derive/src/to_client.rs
```diff
@@ -0,0 +1,134 @@
...
+                        let result = serde_json::from_value::<#returns>(value)
+                            .map_err(|_| Error::new(ErrorCode::ParseError));
+                        match result {
+                            Ok(result) => future::ok(result),
+                            Err(error) => future::err(error),
+                        }
+                    }
+                })
+            }
+            _ => {}
```
tomusdrw: I'd error out or at least print a warning in case of subscriptions if we generate the client, so that people know it's not supported.

derive/src/rpc_trait.rs
dvc94ch: To support generics we also need information about the implementations, to figure out the concrete types. I think it's probably a good idea to create a json file of the required information for generating a client as an intermediate step. Then generating the client would be optional by using something like `generate_client_from_json!()`.

dvc94ch: Now that we are using a struct for the client handling generics is quite easy, ignore my previous comment :)

dvc94ch:
So this is all the glue needed to get a working substrate api client. I'll open a PR there, still need to finish some items on this PR first.

```rust
//! Substrate API
//!
//! Provides a client for a substrate node parameterized over it's
//! transaction pool type.
#![deny(missing_docs)]
#![deny(warnings)]
use jsonrpc_core::client::RpcChannel;
use sr_primitives::generic::SignedBlock;
use sr_primitives::traits;
use substrate_transaction_pool::txpool::{
    ChainApi as PoolChainApi,
};

mod author {
    use substrate_rpc::author::gen_client::Client;
    use substrate_transaction_pool::txpool::{
        BlockHash,
        ChainApi as PoolChainApi,
        ExHash,
    };
    #[allow(type_alias_bounds)]
    pub type AuthorApiClient<P: PoolChainApi> =
        Client<ExHash<P>, BlockHash<P>>;
}

mod chain {
    use substrate_rpc::chain::gen_client::Client;
    use sr_primitives::generic::SignedBlock;
    use sr_primitives::traits::{self, NumberFor};
    #[allow(type_alias_bounds)]
    pub type ChainApiClient<B: traits::Block> =
        Client<NumberFor<B>, B::Hash, B::Header, SignedBlock<B>>;
}

mod state {
    use substrate_rpc::state::gen_client::Client;
    use sr_primitives::traits;
    #[allow(type_alias_bounds)]
    pub type StateApiClient<B: traits::Block> =
        Client<B::Hash>;
}

mod system {
    use substrate_rpc::system::gen_client::Client;
    use sr_primitives::traits::{self, Header};
    #[allow(type_alias_bounds)]
    pub type SystemApiClient<B: traits::Block> =
        Client<B::Hash, <B::Header as Header>::Number>;
}

/// The Substrate API client.
#[derive(Clone)]
pub struct Client<P: PoolChainApi> {
    /// Authoring API
    pub author: author::AuthorApiClient<P>,
    /// Blockchain API
    pub chain: chain::ChainApiClient<P::Block>,
    /// State API
    pub state: state::StateApiClient<P::Block>,
    /// System API
    pub system: system::SystemApiClient<P::Block>,
}

impl<P: PoolChainApi> Client<P>
where
    P::Hash: serde::de::DeserializeOwned,
    SignedBlock<P::Block>: serde::de::DeserializeOwned,
    <P::Block as traits::Block>::Header: serde::de::DeserializeOwned,
{
    /// Creates a new Substrate client.
    pub fn new(channel: RpcChannel) -> Self {
        Client {
            author: author::AuthorApiClient::<P>::new(channel.clone()),
            chain: chain::ChainApiClient::<P::Block>::new(channel.clone()),
            state: state::StateApiClient::<P::Block>::new(channel.clone()),
            system: system::SystemApiClient::<P::Block>::new(channel),
        }
    }
}

impl<P: PoolChainApi> From<RpcChannel> for Client<P>
where
    P::Hash: serde::de::DeserializeOwned,
    SignedBlock<P::Block>: serde::de::DeserializeOwned,
    <P::Block as traits::Block>::Header: serde::de::DeserializeOwned,
{
    fn from(channel: RpcChannel) -> Self {
        Client::new(channel)
    }
}
```

derive/examples/std.rs
dvc94ch: If the return type ends with Result it will take the first generic type argument as the returns type.

derive/src/rpc_trait.rs
dvc94ch: Do you prefer opt-in or opt-out?

core/src/client.rs
```diff
@@ -0,0 +1,231 @@
...
+		}
+	}
+}
+
+impl Future for RpcFuture {
+	type Item = Value;
+	type Error = RpcError;
+
+	fn poll(&mut self) -> Result<Async<Self::Item>, Self::Error> {
+		// TODO should timeout
```
tomusdrw: This can be implemented as a wrapper, do you mind creating an issue for this and assigning a number here?

core/src/client.rs
```diff
@@ -0,0 +1,231 @@
...
+			Ok(Async::Ready(Err(error))) => Err(RpcError::JsonRpcError(error)),
+			Ok(Async::NotReady) => Ok(Async::NotReady),
+			Err(error) => Err(RpcError::Other(error.into())),
+		}
+	}
+}
+
+/// A message sent to the `RpcClient`. This is public so that
+/// the derive crate can generate a client.
+pub struct RpcMessage {
```
tomusdrw: Shall we move the entire client to separate crate? I'd prefer to keep `core` as lean as possible

core/src/client.rs
```diff
@@ -0,0 +1,231 @@
...
+		loop {
+			let msg = match self.ch.poll() {
+				Ok(Async::Ready(Some(msg))) => msg,
+				Ok(Async::Ready(None)) => {
+					// When the channel is dropped we still need to finish
+					// outstanding requests.
+					break;
+				},
+				Ok(Async::NotReady) => break,
+				Err(()) => continue,
```
tomusdrw: Won't we end up in a an endless loop if `Sender` is dropped?

core/src/client.rs
```diff
@@ -0,0 +1,231 @@
...
+			let request = Request::Single(Call::MethodCall(MethodCall {
+				jsonrpc: Some(Version::V2),
+				method: msg.method,
+				params: msg.params,
+				id: id.clone(),
+			}));
+			self.queue.insert(id, msg.sender);
+			let request_str = serde_json::to_string(&request)
+				.map_err(|error| RpcError::Other(error.into()))?;
+			self.sink.start_send(request_str)?;
```
tomusdrw: This is not handled correctly - the `start_send` may return `Ok(NotReady(Requested))` in case the stream is full. In such case the msg should be retried, see: https://docs.rs/futures/0.1.7/futures/enum.AsyncSink.html

core/src/client.rs
```diff
@@ -0,0 +1,231 @@
...
+					return Ok(Async::Ready(()))
+				},
+				Ok(Async::NotReady) => break,
+				Err(err) => Err(err)?,
+			};
+			let response = serde_json::from_str::<Response>(&response_str)
+				.map_err(|error| RpcError::Other(error.into()))?;
+			let outputs: Vec<Output> = match response {
+				Response::Single(output) => vec![output],
+				Response::Batch(outputs) => outputs,
```
tomusdrw: We never send a `Batch` request so this branch should just trigger a warning an be ignored.

core/src/client.rs
```diff
@@ -0,0 +1,231 @@
...
+
+impl<TSink, TStream> Future for RpcClient<TSink, TStream>
+where
+	TSink: Sink<SinkItem=String, SinkError=RpcError>,
+	TStream: Stream<Item=String, Error=RpcError>,
+{
+	type Item = ();
+	type Error = RpcError;
+
+	fn poll(&mut self) -> Result<Async<Self::Item>, Self::Error> {
```
tomusdrw: The implementation seems correct, but it's a bit tricky to get it right. Couldn't we have two separate futures with shared state?
1. The first one would be super simple, cause we just pipe the Sink and Stream together and convert the requests from client into strings to send
2. The second one would handle responses receiving.

To avoid Mutexes we could also have a channel from 1->2 so that the `queue` is local to the second future.

derive/examples/generic-trait-bounds.rs
```diff
@@ -16,23 +16,23 @@ pub trait Rpc<One, Two, Three>
...
-	fn call(&self, _: One) -> FutureResult<(One, u64), Error>;
+	fn call(&self, a: One) -> FutureResult<(One, u64), Error>;
 }
 
 struct RpcImpl;
 
 #[derive(Serialize, Deserialize)]
 struct InAndOut { foo: u64 }
-#[derive(Deserialize)]
+#[derive(Serialize, Deserialize)]
```
tomusdrw: This should be only required when client is generated. Could you please alter this example to not do that so that we are sure that generic parameters bounds are relaxed?

Maybe create another example that uses the client and requires the parameters to be ser-de

derive/src/to_delegate.rs
```diff
@@ -390,11 +393,20 @@ fn generate_where_clause_serialization_predicates(item_trait: &syn::ItemTrait) -
...
 				parse_quote!(Send + Sync + 'static);
 			// add json serialization trait bounds
-			if visitor.serialize_type_params.contains(&ty.ident) {
-				bounds.push(parse_quote!(_serde::Serialize))
-			}
-			if visitor.deserialize_type_params.contains(&ty.ident) {
-				bounds.push(parse_quote!(_serde::de::DeserializeOwned))
+			if client {
+				if visitor.serialize_type_params.contains(&ty.ident) {
+					bounds.push(parse_quote!(_serde::de::DeserializeOwned))
```
tomusdrw: Don't we need both bounds?

ws/client/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "jsonrpc-ws-client"
+version = "0.1.0"
+authors = ["David Craven <david@craven.ch>"]
+edition = "2018"
+
+[dependencies]
+failure = "0.1"
+futures = "0.1"
+jsonrpc-core = { version = "*", path = "../../core" }
```
tomusdrw: Please specify a verison here, otherwise we won;'t be able to push it to crates.io

ws/client/src/lib.rs
```diff
@@ -0,0 +1,37 @@
+use failure::{Error, format_err};
+use futures::prelude::*;
+use futures::sync::mpsc;
+use websocket::{ClientBuilder, OwnedMessage};
+use jsonrpc_core::client::{RpcClient, RpcChannel, RpcError};
+
+pub fn connect<T>(url: &str) -> Result<impl Future<Item=T, Error=RpcError>, Error>
```
tomusdrw: Missing docs, please add `#![warn(missing_docs)]` as well.

ws/client/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+			let (sink, stream) = client
+				.map_err(|err| RpcError::Other(err.into()))
+				.sink_map_err(|err| RpcError::Other(err.into()))
+				.with(|data| -> Result<_, RpcError> {
+					Ok(OwnedMessage::Text(data))
+				})
+				.and_then(|recv| {
+					match recv {
+						OwnedMessage::Text(data) => Ok(data),
+						// TODO: pings and pongs and close messages need to be
```
tomusdrw: Is there anything that prevents the implementation? Can we log this as a follow up issue?

ws/client/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+						//       answered; for now we produce an error when that
+						//       happens
+						_ => Err(RpcError::Other(format_err!("unimplemented"))),
+					}
+				}).split();
+			let (sender, receiver) = mpsc::channel(0);
```
tomusdrw: I guess unbufferred channel here is fine, but that's something to mention in docs.

core/src/client.rs
dvc94ch: Can't a different jsonrpc server implementation return a batch response?

derive/src/to_delegate.rs
dvc94ch: I don't think so, that would cause a compilation error of the generated code. If it is a parameter type in the server that needs to be serialized it's a result type in the client that needs to be deserialized and vice versa. The only reason it would need both bounds is if the same type is used as a parameter and a return type, which is handled correctly.

core/src/client.rs
tomusdrw: The server should only respond with `Batch` if client sends a batch request: https://www.jsonrpc.org/specification#batch

If you prefer to leave as is I'm also fine, but I think not handling `Batch` would be a bit more strict and would simplify the code tiny bit.

derive/src/to_delegate.rs
tomusdrw: Fair enough, although:

> If it is a parameter type in the server that needs to be serialized it's a result type in the client that needs to be deserialized and vice versa.

This implies that the type is Ser+De anyway, but I guess relaxed bounds just on the client interface are a good thing.

dvc94ch:
Changes are in `Implement jsonrpc client.`, `Make client/server generation optional.` and `Add websocket client.`

client/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+authors = ["Parity Technologies <admin@parity.io>"]
+description = "Transport agnostic JSON-RPC 2.0 client implementation."
+documentation = "https://docs.rs/jsonrpc-core/"
```
tomusdrw:
```suggestion
documentation = "https://docs.rs/jsonrpc-client/"
```

core/src/types/error.rs
```diff
@@ -157,3 +157,11 @@ impl Error {
...
 	}
 }
+
+impl std::fmt::Display for Error {
+	fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+		write!(f, "{}: {}", self.code.description(), self.message)
+	}
+}
+
+impl std::error::Error for Error {}
```
tomusdrw: Closes #253

ws/client/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+authors = ["Parity Technologies <admin@parity.io>"]
+description = "JSON-RPC 2.0 websocket client implementation."
+documentation = "https://docs.rs/jsonrpc-core/"
```
tomusdrw:
```suggestion
documentation = "https://docs.rs/jsonrpc-ws-client/"
```

client/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+	type Item = ();
+	type Error = RpcError;
+
+	fn poll(&mut self) -> Result<Async<Self::Item>, Self::Error> {
+		// Handle requests from the client.
+		loop {
+			if self.channel.is_none() {
+				break;
+			}
+			let msg = match self.channel.as_mut().unwrap().poll() {
```
ascjones: We prefer to use `expect("<proof>; qed")` instead of `unwrap`s. So something like `expect("channel is some; qed")`.

derive/src/rpc_trait.rs
```diff
@@ -156,12 +158,15 @@ fn generate_rpc_item_trait(item_trait: &syn::ItemTrait) -> Result<(syn::ItemTrai
 		rpc_trait.has_metadata,
 		has_pubsub_methods,
 	)?;
+
+	let rpc_client_module = generate_client_module(&method_registrations, &item_trait)?;
```
ascjones: Could we pass in `DeriveOptions` and only generate the client/server modules where necessary? If the impact on compile time is negligible then it should be fine. But we have some pretty big rpc traits e.g. in `parity-ethereum`.

derive/src/to_client.rs
```diff
@@ -0,0 +1,280 @@
...
+		pub mod gen_client {
+			use super::*;
+			use _jsonrpc_core::{
+				Call, Error, ErrorCode, Id, MethodCall, Params, Request,
+				Response, Version,
+			};
+			use _jsonrpc_core::futures::{future, Future, Sink};
+			use _jsonrpc_core::futures::sync::oneshot;
+			use _jsonrpc_core::serde_json::{self, Value};
+			use jsonrpc_client::{RpcChannel, RpcError, RpcFuture, RpcMessage};
```
ascjones: Could we use `proc_macro_crate::crate_name("jsonrpc-client")` to be consistent with other imports and protect against crate renames

tomusdrw:
:tada:

bkchr:
First, thanks for doing this!

Second, sorry for coming late to the party, but seen from a use perspective having the client in `gen_client` is somewhat weird? Maybe we could support providing a custom name for the generated client. 
I also have seen that you needed to replace `_` with actual names. I understand why you have done this, but again, from a use perspective it is not that nice. If you look [here](https://github.com/paritytech/substrate/blob/master/core/sr-api-macros/src/utils.rs#L105), you can see how this is solved in substrate. We just replace `_` with some unique name in the code generation :)

tomusdrw:
@bkchr let's maybe create a separate issues for this, the PR was getting really big already:
https://github.com/paritytech/jsonrpc/issues/414

bkchr:
Yeah good point, I was a little bit to lazy to create an issue :sweat_smile:

PR #407, opened 2019-03-27
Use rustfmt

### LINKED ISSUE #406: Use rustfmt
There seems to be an option in rustfmt called `hard_tabs`.

`rustfmt.toml`
```toml
hard_tabs = true
```

dvc94ch:
The PR #405 is nearing completion. Didn't get a comment on #406, so I thought I'd open a PR. I'll rebase #405 and split it into manageable chunks for review.

Closes #406

### DIFF
--- a/derive/tests/macros.rs
+++ b/derive/tests/macros.rs
@@ -1,6 +1,6 @@
-use serde_json;
-use jsonrpc_derive::rpc;
 use jsonrpc_core::{IoHandler, Response};
+use jsonrpc_derive::rpc;
+use serde_json;
 
 pub enum MyError {}
 impl From<MyError> for jsonrpc_core::Error {
@@ -88,11 +88,17 @@ fn should_accept_single_param() {
 
 	// then
 	let result1: Response = serde_json::from_str(&res1.unwrap()).unwrap();
-	assert_eq!(result1, serde_json::from_str(r#"{
+	assert_eq!(
+		result1,
+		serde_json::from_str(
+			r#"{
 		"jsonrpc": "2.0",
 		"result": -1,
 		"id": 1
-	}"#).unwrap());
+	}"#
+		)
+		.unwrap()
+	);
 }
 
 #[test]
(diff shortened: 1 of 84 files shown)


dvc94ch:
@tomusdrw This PR screws with contributor ranking for people who care about these things. I think it would be better if you run `cargo fmt` if you decide it improved the code formatting.

derive/tests/macros.rs
```diff
@@ -130,17 +142,28 @@ fn should_use_method_name_aliases() {
 
 	// then
 	let result1: Response = serde_json::from_str(&res1.unwrap()).unwrap();
-	assert_eq!(result1, serde_json::from_str(r#"{
+	assert_eq!(
+		result1,
+		serde_json::from_str(
+			r#"{
```
tomusdrw: That looks a bit ugly :P

tomusdrw:
Looks good to me, let's give it a shot. Can you fix the outstanding issues?

derive/tests/macros.rs
dvc94ch: Do you want me to manually format this? It kind of defeats the point of using rustfmt and will cause errors in the ci.

tomusdrw: No, not at all. Just a nitpick, that rustfmt is not perfect, perhaps we can tune some options in the future to get rid of that.
In the other comment I meant the errors that causes the CI to fail. If CI is green I'm good to merge this.

dvc94ch: The error is a formatting issue on stable where it expects a semicolon but on beta/nightly it doesn't. rustfmt 1.0 (with default settings) is supposed to be formatting compatible between versions unless it's a bug. It is a bit unfortunate, but not really sure what to do.

tomusdrw: Let's run `rustfmt` only on stable then. We've been through some issues like this in the past in some other repo afair (after all we killed rustfmt completely).

tomusdrw:
Thanks!

PR #408, opened 2019-03-29
access to websocket broadcaster

jacobrosenthal:
Im sending out a broadcaster object holding the ws::sender in order to not leak ws stuff which seems to work, but for now Im only accepting vec u8 for send, instead of From Message Any thoughts on how to implement that from without reexporting Message?

### DIFF
--- a/ws/src/server.rs
+++ b/ws/src/server.rs
@@ -38,6 +38,13 @@ impl Server {
 		&self.addr
 	}
 
+	/// Returns a Broadcaster that can be used to send messages on all connections.
+	pub fn broadcaster(&self) -> Broadcaster {
+		Broadcaster {
+			broadcaster: self.broadcaster.clone(),
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


jacobrosenthal:
[clabot:check]

jacobrosenthal:
Well. That seemed to work just fine. Thoughts on the implementation?

ws/src/server.rs
```diff
@@ -16,138 +16,181 @@ use crate::session;
 
 /// `WebSockets` server implementation.
 pub struct Server {
-	addr: SocketAddr,
-	handle: Option<thread::JoinHandle<Result<()>>>,
-	executor: Arc<Mutex<Option<Executor>>>,
-	broadcaster: ws::Sender,
+    addr: SocketAddr,
```
tomusdrw: We use tabs across the codebase please re-format or wait for #406

tomusdrw:
Awesome, thanks a lot!

PR #413, opened 2019-04-02
Bump version (11.0)

ascjones:
Major release: 
 - macros removed 
 - client support added.


bkchr:
I would propose to mention the client support somewhere and it would be nice to have a prominent example as well. Otherwise the users will never know that it exists :)

derive/src/lib.rs
```diff
@@ -123,6 +123,38 @@
...
+//! 	fn protocol_version(&self) -> Result<String>;
+//! 
+//! 	//! / Adds two numbers and returns a result
+//! 	#[rpc(name = "add", alias("callAsyncMetaAlias"))]
+//! 	fn add(&self, a: u64, b: u64) -> Result<u64>;
+//! }
+//! 
+//! fn main() {
+//! 	let fut = {
+//! 		let (sender, _) = mpsc::channel(0);
```
tomusdrw: I'd add a comment what the receiver is for (or even spawn a new thread that actually responds).

ascjones: I've updated it to use the client/server example (copied from `derive/examples/std`)

PR #417, opened 2019-04-04
Extract client plumbing from derive, fix circular dependency

ascjones:
A circular dev dependency in `core-client` was preventing the release of this crate. A test was using client `derive` which requires a dependency on `core-client`

This PR removes generated client from `derive` and uses a handrolled client. For this I extracted some code from `derive` which handles the dispatch and serialization of messages into `RawClient` and `TypedClient` helpers.

### DIFF
--- a/core-client/src/lib.rs
+++ b/core-client/src/lib.rs
@@ -259,22 +259,25 @@ impl TypedClient {
 		let params = match args {
 			Value::Array(vec) => Params::Array(vec),
 			Value::Null => Params::None,
-			_ => panic!("RPC params should serialize to a JSON array, or null"),
+			_ => return future::Either::A(future::err(RpcError::Other(
+				format_err!("RPC params should serialize to a JSON array, or null")))),
 		};
 
-		self.0
-			.call_method(method, params)
-			.and_then(move |value: Value| {
-				log::debug!("response: {:?}", value);
-				let result = serde_json::from_value::<R>(value)
-					.map_err(|error| {
-						RpcError::ParseError(
-							returns.into(),
-							error.into(),
-						)
-					});
-				future::done(result)
-			})
+		future::Either::B(
+			self.0
+				.call_method(method, params)
+				.and_then(move |value: Value| {
+					log::debug!("response: {:?}", value);
+					let result = serde_json::from_value::<R>(value)
+						.map_err(|error| {
+							RpcError::ParseError(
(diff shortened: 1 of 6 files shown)


core-client/src/lib.rs
```diff
@@ -202,6 +204,80 @@ where
...
+		method: &str,
+		returns: &'static str,
+		args: T,
+	) -> impl Future<Item=R, Error=RpcError> {
+		let args = serde_json::to_value(args)
+			.expect("Only types with infallible serialisation can be used for JSON-RPC");
+		let params = match args {
+			Value::Array(vec) => Params::Array(vec),
+			Value::Null => Params::None,
+			_ => panic!("RPC params should serialize to a JSON array, or null"),
```
ascjones: Wasn't sure what to do here. This was not possible when this was in `derive` because the params are guaranteed to be a Tuple.

tomusdrw: Maybe just return an `RpcError`?

core-client/src/lib.rs
```diff
@@ -284,12 +360,12 @@ pub mod local {
...
 mod tests {
-	use futures::prelude::*;
-	use jsonrpc_core_client::local;
+	use super::*;
 	use jsonrpc_core::{IoHandler, Result};
 	use jsonrpc_derive::rpc;
+	use crate::{TypedClient, RpcError, RpcChannel};
 
-	#[rpc]
+	#[rpc(server)]
```
tomusdrw: Would be good to actually add some tests for the generated client somewhere too. Maybe in a separate crate?

core-client/src/lib.rs
ascjones: Yes I did that originally before changing to panic, I think that might be better after all.

dvc94ch:
Nice! You can also remove the log dependency from the derive crate, since it's not needed anymore. And the `extern crate log` from the compile tests.

dvc94ch:
We can also make RpcMessage private now.

ascjones:
@dvc94ch making it fully private was not straightforward since `RpcMessage` appears in `RpcClient::new` which is used in the `ws` crate. Would require some refactoring there.

I have however made the fields private.

Let me know if that all looks good and I will merge this and get `11.0` released to crates.io!

dvc94ch:
@ascjones lgtm

PR #420, opened 2019-04-16
Client http transport

ascjones:
Rel: https://github.com/paritytech/substrate/pull/2255. 

Still todo:
- [x] Improve ergonomics of channel impl - helper for the transport Future
   - Use a combinator e.g. `join` or wrap in a struct which impls `Future`
- [x] Tests/examples of http client in use
- [x] Fix error handling in core client, channels shouldn't be dropped because of rpc errors


ascjones:
I've changed my approach here: initially I copied @tomusdrw's approach of creating a pair of channels for the `Sink/Stream` and creating the client with those. 

https://github.com/paritytech/jsonrpc/blob/7a1c1f487387e4cfc25778baa0a33223bf6abf8a/core-client/src/transports/http.rs#L17-L33
https://github.com/paritytech/jsonrpc/blob/7a1c1f487387e4cfc25778baa0a33223bf6abf8a/core-client/src/transports/mod.rs#L10-L40

However this ran into problems where HTTP errors would not be returned to the caller, because there is no way to correlate the error response with the sender in  `RpcClient`, since the request `Id` is only available in a successful response from the server.

So I've created a "standalone" HTTP transport, which can handle the errors properly since it is request/response instead of duplex.

I've extracted the former `RpcClient` into `transports` and renamed it `Duplex`. Shared functionality between that and the HTTP transport (id assignment, request building. response handling) I've extracted out into some helpers.

Still a bit more work to do on this - but would be good to have some feedback to see whether this is the right direction.

dvc94ch:
> However this ran into problems where HTTP errors would not be returned to the caller, because there is no way to correlate the error response with the sender in RpcClient, since the request Id is only available in a successful response from the server.

What happens currently? Is the error at least logged?

ascjones:
I think it would only be logged if you attach a logger to the client future error e.g. https://github.com/paritytech/substrate/pull/2255/files#diff-74c6a3e10a271935aa30a9981463e681R76

core-client/Cargo.toml
```diff
@@ -20,15 +20,19 @@ categories = [
 
 [dependencies]
 failure = "0.1"
-futures = "~0.1.6"
+futures = "0.1.26"
+hyper = "0.12"
```
tomusdrw: Should we make the transports features? Maybe worth a separate issue.

core-client/src/lib.rs
```diff
@@ -38,168 +40,60 @@ impl From<Error> for RpcError {
...
-	pub fn new(sink: TSink, stream: TStream, channel: mpsc::Receiver<RpcMessage>) -> Self {
-		RpcClient {
-			id: 0,
-			queue: HashMap::new(),
-			sink,
-			stream,
-			channel: Some(channel),
-			outgoing: VecDeque::new(),
-		}
+impl Into<RpcChannel> for mpsc::Sender<RpcMessage> {
```
tomusdrw: I think we should rather implement `From`:
> Library authors should not directly implement this[Into] trait, but should prefer implementing the From trait, which offers greater flexibility and provides an equivalent Into implementation for free, thanks to a blanket implementation in the standard library.

core-client/src/transports/duplex.rs
```diff
@@ -0,0 +1,126 @@
...
+			queue: HashMap::new(),
+			sink,
+			stream,
+			channel: Some(channel),
+			outgoing: VecDeque::new(),
+		}
+	}
+
+	/// Creates a new `Duplex`, along with a channel to communicate
+	pub fn with_channel(sink: TSink, stream: TStream) -> (Self, RpcChannel) {
```
tomusdrw: The method is a bit confusing, but seeing `with_channel` I think I was expecting the channel to be passed as parameter instead, I'd swap the names.

core-client/src/transports/duplex.rs
```diff
@@ -0,0 +1,126 @@
...
+	type Item = ();
+	type Error = RpcError;
+
+	fn poll(&mut self) -> Result<Async<Self::Item>, Self::Error> {
+		// Handle requests from the client.
+		loop {
+			if self.channel.is_none() {
+				break;
+			}
+			let msg = match self.channel.as_mut().expect("channel is some; qed").poll() {
```
tomusdrw: You could match on `self.channel.as_mut()` to avoid `expect`

core-client/src/transports/duplex.rs
```diff
@@ -0,0 +1,126 @@
...
+				Ok(Async::Ready(Some(msg))) => msg,
+				Ok(Async::Ready(None)) => {
+					// When the channel is dropped we still need to finish
+					// outstanding requests.
+					self.channel.take();
+					break;
+				}
+				Ok(Async::NotReady) => break,
+				Err(()) => continue,
```
tomusdrw: That will lead to an endless loop, why do we `continue` if the `Sender` is dropped? We should rather resolve with an error after processing the `outgoing` queue?

core-client/src/transports/duplex.rs
```diff
@@ -0,0 +1,126 @@
...
+			};
+			let (id, request_str) = self.request_builder.single_request(&msg);
+			self.queue.insert(id, msg.sender);
+			self.outgoing.push_back(request_str);
+		}
+		// Handle outgoing rpc requests.
+		loop {
+			match self.outgoing.pop_front() {
+				Some(request) => match self.sink.start_send(request)? {
```
tomusdrw: Wouldn't it be easier to use `send_all` instead?

core-client/src/transports/duplex.rs
```diff
@@ -0,0 +1,126 @@
...
+			};
+
+			let responses = super::parse_response(&response_str)?;
+			for (id, result) in responses {
+				let channel = self.queue.remove(&id);
+				match channel {
+					Some(tx) => tx
+						.send(result)
+						.map_err(|_| RpcError::Other(format_err!("oneshot channel closed")))?,
+					None => Err(RpcError::UnknownId)?,
```
tomusdrw: I don't think it should be an error - this means the server can terminate the entire stream by sending a malformed response. I'd rather handle that gracefuly with a warning and continue

core-client/src/transports/duplex.rs
```diff
@@ -0,0 +1,126 @@
...
+				};
+			}
+		}
+		if self.channel.is_none() && self.outgoing.is_empty() && self.queue.is_empty() && done_sending {
+			debug!("client finished");
+			Ok(Async::Ready(()))
+		} else {
+			Ok(Async::NotReady)
```
tomusdrw: I'm not sure if this is correct, are we sure that in every case where we go into this branch we `poll` something upstream? Otherwise the task is not going to be woken up, I was looking at different cases, but it seems correct, so just leaving a notice here :)

core-client/src/transports/http.rs
```diff
@@ -0,0 +1,308 @@
...
+	let fut = receiver
+		.map(move |msg: RpcMessage| {
+			let (_, request) = request_builder.single_request(&msg);
+			let request = Request::post(&url)
+				.header(
+					http::header::CONTENT_TYPE,
+					http::header::HeaderValue::from_static("application/json"),
+				)
+				.body(request.into())
+				.unwrap();
```
tomusdrw: could do with a proof

core-client/src/transports/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+		.map(|response| {
+			let outputs: Vec<Output> = match response {
+				Response::Single(output) => vec![output],
+				Response::Batch(outputs) => outputs,
+			};
+			outputs
+				.into_iter()
+				.map(|output| {
+					let id = output.id().clone();
+					let value: Result<Value, Error> = output.into();
```
tomusdrw: This conversion here might be a bit wasteful, can't we have some more strict type than `Value`?

core-client/Cargo.toml
ascjones: https://github.com/paritytech/jsonrpc/issues/431

core-client/src/transports/mod.rs
ascjones: Wasteful in what way? The `output` contains the json `Value` and it is just transformed to a `Result` https://github.com/paritytech/jsonrpc/blob/31ec6d67f2ab338a26f5080af5804960f7ab39e4/core/src/types/response.rs#L78

core-client/src/transports/duplex.rs
ascjones: I've pulled this out to a public function `duplex`. Not sure about the name but perhaps more idiomatic for creating the `(Duplex, RpcChannel)` tuple.

core-client/src/transports/duplex.rs
ascjones: Seems the futures impl of mpsc [returns](https://docs.rs/futures/0.1.27/src/futures/sync/mpsc/mod.rs.html#945) `Ready(None)` when the channel is closed - which is handled above.

And indeed that is [documented](https://docs.rs/futures/0.1.27/futures/stream/trait.Stream.html#errors) as the expected behaviour for `Stream`, and for [future versions](https://github.com/rust-lang-nursery/futures-rs/issues/206).

ascjones:
I've opened an issue to address some of the points with `duplex`. I think those changes will be better as a separate PR since I've just moved/renamed that code in this PR. https://github.com/paritytech/jsonrpc/issues/432

core-client/src/transports/duplex.rs
ascjones: Task in https://github.com/paritytech/jsonrpc/issues/432

core-client/src/transports/duplex.rs
ascjones: Task in https://github.com/paritytech/jsonrpc/issues/432

core-client/src/transports/duplex.rs
ascjones: Task in https://github.com/paritytech/jsonrpc/issues/432

PR #421, opened 2019-04-16
fix(tcp poll): don't drop when receiver value is ready

niklasad1:
Basically when `up` is polled and the result is `Ready(None)` we need to check if `receiver` has received any result which this commit fixes.

In other words the current implementation will return `early` and then drop it and some result in the `receiver` could be missed!

I tried to mimic to previous implementation to prioritize `up` and used pattern matching on both at the same time for verbosity!

### DIFF
--- a/tcp/src/dispatch.rs
+++ b/tcp/src/dispatch.rs
@@ -12,7 +12,7 @@ pub type SenderChannels = Mutex<HashMap<SocketAddr, mpsc::Sender<String>>>;
 
 pub struct PeerMessageQueue<S: Stream> {
 	up: S,
-	receiver: mpsc::Receiver<String>,
+	receiver: Option<mpsc::Receiver<String>>,
 	_addr: SocketAddr,
 }
 
@@ -20,7 +20,7 @@ impl<S: Stream> PeerMessageQueue<S> {
 	pub fn new(response_stream: S, receiver: mpsc::Receiver<String>, addr: SocketAddr) -> Self {
 		PeerMessageQueue {
 			up: response_stream,
-			receiver,
+			receiver: Some(receiver),
 			_addr: addr,
 		}
 	}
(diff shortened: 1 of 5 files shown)


tcp/src/dispatch.rs
```diff
@@ -84,23 +84,18 @@ impl<S: Stream<Item = String, Error = std::io::Error>> Stream for PeerMessageQue
...
-				return Ok(Async::Ready(None));
-			}
-			_ => {}
-		}
 
-		match self.receiver.poll() {
-			Ok(result) => Ok(result),
-			Err(send_err) => {
+		match (self.up.poll(), self.receiver.poll()) {
+			(Ok(Async::Ready(Some(v))), _) => Ok(Async::Ready(Some(v))),
```
tomusdrw: I'm not 100% sure about this, it seems that we might loose some items from `receiver` stream if both are ready.
Perhaps we could introduce simple buffering by introducing an `Option<Poll<...>>` that is checked before polling.

bkchr: Sounds like we want to use this: https://docs.rs/futures/0.1.26/futures/stream/struct.Select.html

niklasad1: yeah, I think so the only problem is that the error types, `() vs std::io::Error`

tomusdrw: one of the errors is ignored anyway (or "should never happen"), so we can just map it to the other.

niklasad1: I haven't been able to use `Stream::select` successfully in this case because it requires both streams != NotReady, https://docs.rs/futures/0.1.26/src/futures/stream/select.rs.html#35-61

When I have manually tested it, one channel is `NotReady` and the other one is `active`.
Thus, I'm not sure how to solve it properly. 

A potential `work-around` could be drop it when `ch1 == NotReady && ch2 == Ready(None)` but we then could potentially loose data if ch1 receive data in the future.

Thoughts?

bkchr: I don't understand why both need to be `NotReady`?

niklasad1: Sorry for the bad/faulty explanation, I meant that both streams need to return `Ready(None)` for it to be resolved

bkchr: Ahh, I thought the whole purpose of this pr was to fix exactly this :D Then, `Stream::select` is not correct. 
So, we need to look if `receiver` received some messages but don't wait for it to end? Couldn't it be that we get to a point where `receiver` always returns a value, while `up` is already finished? 
That sounds like some weird behavior.

niklasad1: true true I agree that we ideally we should use `Stream::select`, guess I got frustrated yesterday let me debug this further

PR #424, opened 2019-04-30
Be forward compatible with rust-lang/rust#59928

Centril:
Hello! In https://github.com/rust-lang/rust/pull/59928 we are making https://github.com/rust-lang/rust/issues/57644 a deny-by-default lint. To be forward compatible with that, here's a simple fix. Thank you for you understanding!


Centril:
(Tho it seems you may have already handled this?)

Issue #433, opened 2019-05-21
TLS support for the HTTP client


abustany:
Use hyper-tls to allow connecting a client over HTTPS. #431 is probably relevant here too, to avoid pulling hyper-tls unconditionally.

ascjones:
e.g. https://github.com/tomusdrw/rust-web3/blob/master/src/transports/http.rs#L97

dvc94ch:
Closed by #435

PR #435, opened 2019-05-22
Optional TLS support for HTTP transport (#433)

abustany:


abustany:
[clabot:check]

abustany:
[clabot:check]

abustany:
I'm not sure how to add a test for this... Testing all features will ensure that the `tls` feature does not break http, but do we have an easy to spin up an actual https server? It looks like jsonrpc-http-server has no helpers for that.

tomusdrw:
@abustany can you add a job to the CI to check that the client compiles with that feature enabled?

abustany:
Added using the build matrix feature from Travis, let me know if you'd prefer using the build script directly instead. I thought using the build matrix would lead to clearer reporting...

abustany:
oh hmmm I just realized the build probably now publishes the documentation twice :-/

abustany:
The fix I pushed is a bit ad-hoc... I don't know if there's a more generic way to implement it.

tomusdrw:
Thanks!

PR #437, opened 2019-05-25
Add a way to close the http server while waiting

dvdplm:
Ref. https://github.com/paritytech/jsonrpc/pull/331

This PR implements the idea put forward by @tomusdrw to add an additional `oneshot` channel to let the `Server` know when the future is shut down, which lets us write an implementation of `Server::wait()` that can be interrupted by an external signal.

(Also includes a few fixed compiler warnings)

### DIFF
--- a/http/src/lib.rs
+++ b/http/src/lib.rs
@@ -437,19 +437,17 @@ impl<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>> ServerBuilder<M, S> {
 			.collect::<io::Result<(Vec<_>)>>()?;
 		handles.push((eloop, close, done_rx));
 
-		let (executors, closers, done_rxs) = handles
+		let (executors, done_rxs) = handles
 			.into_iter()
-			.fold((vec![], vec![], vec![]), |mut acc, (eloop, closer, done_rx)| {
-				acc.0.push(eloop);
-				acc.1.push(closer);
-				acc.2.push(done_rx);
+			.fold((vec![], vec![]), |mut acc, (eloop, closer, done_rx)| {
+				acc.0.push((eloop, closer));
+				acc.1.push(done_rx);
 				acc
 			});
 
 		Ok(Server {
 			address: local_addr?,
 			executors: Arc::new(Mutex::new(Some(executors))),
-			close: Arc::new(Mutex::new(Some(closers))),
 			done: Some(done_rxs),
 		})
 	}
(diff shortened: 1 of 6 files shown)


dvdplm:
The CI failure is an odd one (and not related to this PR afaict). I could reproduce it once locally using rust stable but never with nighlty. After a `cargo clean` the failure went away. Not sure how to debug it tbh.

http/src/lib.rs
```diff
@@ -416,27 +421,36 @@ impl<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>> ServerBuilder<M, S> {
...
+				acc.1.push(closer);
+				acc.2.push(done_rx);
+				acc
+			});
 
 		Ok(Server {
 			address: local_addr?,
-			executor: Some(executors),
-			close: Some(close),
+			executors: Arc::new( Mutex::new(Some(executors))),
```
tomusdrw:
```suggestion
			executors: Arc::new(Mutex::new(Some(executors))),
```

http/src/lib.rs
```diff
@@ -416,27 +421,36 @@ impl<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>> ServerBuilder<M, S> {
...
+			close: Arc::new(Mutex::new(Some(closers))),
```
tomusdrw: Would it be worth having only one struct under `Arc<Mutex<`?

http/src/lib.rs
```diff
@@ -562,45 +579,67 @@ fn configure_port(_reuse: bool, _tcp: &net2::TcpBuilder) -> io::Result<()> {
...
+#[derive(Clone)]
+pub struct CloseHandle {
+	executors: Arc<Mutex<Option<Vec<Executor>>>>,
+	closers: Arc<Mutex<Option<Vec<oneshot::Sender<()>>>>>,
+}
+
+impl CloseHandle {
+	/// Shutdown a running server
+	pub fn close(self) {
+        if let Some(executors) = self.executors.lock().take() {
```
tomusdrw: Since we always lock both, might be worth introducing a struct for that.

dvdplm: …like you suggested on the other PR! Indeed, you're right.

[This commit](https://github.com/paritytech/jsonrpc/pull/437/commits/b271b119bedf3bf215d381657404d0d9c55ab9dd) uses:

```
#[derive(Clone)]
pub struct CloseHandle(Arc<Mutex<Option<Vec<(Executor, oneshot::Sender<()>)>>>>);
```

dvdplm:
@tomusdrw good to go?

http/src/lib.rs
```diff
@@ -562,45 +577,60 @@ fn configure_port(_reuse: bool, _tcp: &net2::TcpBuilder) -> io::Result<()> {
...
+		self.close_handle().close()
 	}
 
 	/// Will block, waiting for the server to finish.
 	pub fn wait(mut self) {
-		for executor in self.executor.take().expect(PROOF) {
-			executor.wait();
+		if let Some(receivers) = self.done.take() {
+			for receiver in receivers {
+				let _ = receiver.wait();
```
carllin: Is this wait sufficient to guarantee that all the threads have cleaned up? It guarantees that the serving threads have been cleaned up here: https://github.com/paritytech/jsonrpc/blob/master/http/src/lib.rs#L565, but the tokio threads owned by the RpcEventLoop aren't guaranteed to have exited until this `wait` returns from within the `Executor->RpcEventLoop` here: https://github.com/paritytech/jsonrpc/blob/master/server-utils/src/reactor.rs#L75

tomusdrw: I don't think we can wait for the `Executor` to finish - it could have been spawned externally or shared for some other tasks. I guess the issue might be that `wait()` only actually waits for the moment where we stop accepting new requests, not when we stop processing them.
Do you have a specific issue that this behaviour causes for you?

carllin: @tomusdrw, thanks for the reply! I am running into issues writing test code for testing start/ restart behavior on the server. B/c shutting down the server doesn't guarantee the spawned Executor tasks have finished, we are experiencing data races on shared state on restart.

tomusdrw: Please do, but I think it's a tricky one. We need to distinguish between externally provided executor (where we don't really want to wait until it ends) and a one that we've spawned ourselves.

PR #439, opened 2019-05-29
Add client transport features

### LINKED ISSUE #431: Add features for client transports
e.g. `ws` `http`. This way we will not include dependencies that are not required.

rel: https://github.com/paritytech/jsonrpc/pull/420#discussion_r285470134

ascjones:
Resolves #431.

I've changed the CI script to run `core-client` build for the new features with `--manifest-path=core-client/Cargo.toml` since the features don't work with the top level cargo workspace.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,10 @@ before_script:
   - rustup component add rustfmt
 
 script:
-  - cargo build --all --features "$JSONRPC_CLIENT_FEATURES"
-  - cargo test --all --features "$JSONRPC_CLIENT_FEATURES"
+  - cargo build --all
+  - cargo test --all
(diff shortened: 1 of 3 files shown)


dvdplm:
> the features don't work with the top level cargo workspace.

Like in general or just in this project? Seems like a footgun worth knowing about!

tomusdrw:
> Like in general or just in this project? Seems like a footgun worth knowing about!

I think in general, the point of workspace is just to group a bunch of related projects, but a workspace crate can't really be imported or used as a library.

ascjones:
https://github.com/rust-lang/cargo/issues/5015

dvdplm:
Will this fix CI you reckon?

ascjones:
Cleared the caches and reran - seemed to do the trick

PR #446, opened 2019-07-04
Bump version

ascjones:
In preparation for a release to include the pubsub client #436.

### DIFF
--- a/_automate/bump_version.sh
+++ b/_automate/bump_version.sh
@@ -10,9 +10,9 @@ ack "^version = \"" -l | \
 	grep toml | \
 	xargs sed -i "s/^version = \".*/version = \"$VERSION\"/"
 
-ack "{ version = \"$PREV_DEPS" -l | \
+ack "[{|\,] version = \"$PREV_DEPS" -l | \
 	grep toml | \
-	xargs sed -i "s/{ version = \"$PREV_DEPS/{ version = \"$NEW_DEPS/"
+	xargs sed -i "s/\([{|\,]\) version = \"$PREV_DEPS/\1 version = \"$NEW_DEPS/"
 
 ack " = \"$PREV_DEPS" -l | \
     grep md | \
--- a/_automate/bump_version.sh
+++ b/_automate/bump_version.sh
@@ -10,9 +10,9 @@ ack "^version = \"" -l | \
 	grep toml | \
 	xargs sed -i "s/^version = \".*/version = \"$VERSION\"/"
 
-ack "[{|\,] version = \"$PREV_DEPS" -l | \
+ack "{ version = \"$PREV_DEPS" -l | \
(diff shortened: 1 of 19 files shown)


_automate/bump_version.sh
```diff
@@ -10,9 +10,9 @@ ack "^version = \"" -l | \
 	grep toml | \
 	xargs sed -i "s/^version = \".*/version = \"$VERSION\"/"
 
-ack "{ version = \"$PREV_DEPS" -l | \
+ack "[{|\,] version = \"$PREV_DEPS" -l | \
 	grep toml | \
-	xargs sed -i "s/{ version = \"$PREV_DEPS/{ version = \"$NEW_DEPS/"
+	xargs sed -i "s/\([{|\,]\) version = \"$PREV_DEPS/\1 version = \"$NEW_DEPS/"
```
tomusdrw: Maybe better to unify syntax in the `test` crate?

core-client/Cargo.toml
```diff
@@ -8,7 +8,7 @@ keywords = ["jsonrpc", "json-rpc", "json", "rpc", "serde"]
 license = "MIT"
 name = "jsonrpc-core-client"
 repository = "https://github.com/paritytech/jsonrpc"
-version = "12.0.0"
+version = "13.0.0"
```
tomusdrw: Do we really need `13.0`? Can't we just bump to `12.1`? Are there any breaking changes?

dvc94ch: If there is a bug in handling existing #[pubsub] macros it will break builds.

_automate/bump_version.sh
ascjones: Why miss an opportunity to make an unreadable regex? :grin:

core-client/Cargo.toml
tomusdrw: Hmm, but is there?

ascjones: @dvc94ch true but that would be an unintentional "breaking" change, the API itself should still be compatible.

@tomusdrw so `core-client` and `derive` should both go to `12.1` then.?

tomusdrw: I'd rather keep all creates version aligned. So if we feel there is a reason to bump to 13.0 let's do that, but if there are no breaking changes that we know of, let's stick to 12.1 imho.

ascjones: Actually there is also an addition to `core` https://github.com/paritytech/jsonrpc/pull/436/files#diff-77e9ef3aa593be6048a014ed42ef10dbR34

So upgrade all crates to 12.1?

dvc94ch:
So is this happening? :)

ascjones:
Waiting on something from @tomusdrw

PR #447, opened 2019-07-08
Introduce IoDelegateExtension.

tomusdrw:


derive/src/to_delegate.rs
```diff
@@ -121,6 +121,7 @@ pub fn generate_trait_item_method(
 		}
 	};
 
+
```
ascjones: https://en.wikipedia.org/wiki/Whitespace_(programming_language)

PR #448, opened 2019-07-09
Strict and non-strict parsing for `core`

tomusdrw:
Some types in `Response` were missing `deny_unknown_fields` that would cause invalid response to be parsed correctly when `Deserializing`.

However later I started thinking if having a less strict version would be useful as well, hence I introduced the `strict` feature.

I'd like to hear your opinions on whether we want to maintain `strict` and `non-strict` versions or rather default to `strict`.

See: https://github.com/tomusdrw/rust-web3/issues/222#issuecomment-509695268 why non strict mode would be useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,6 @@ before_script:
 script:
   - cargo build --all
   - cargo test --all
-  - cargo test --manifest-path ./core/Cargo.toml --no-default-features
   - |
     ([ $TRAVIS_RUST_VERSION = stable ] && cargo fmt --all -- --check) || true
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ before_script:
 script:
   - cargo build --all
   - cargo test --all
+  - cargo test --manifest-path ./core/Cargo.toml --no-default-features
   - |
     ([ $TRAVIS_RUST_VERSION = stable ] && cargo fmt --all -- --check) || true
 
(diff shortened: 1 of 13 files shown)


debris:
I like `strict` more... jsonrpc provides a way to send arbitrary data in a `params` field, so I don't see a reason why someone would need to use other fields

tomusdrw:
@debris Yeah, I'm more inclined to go with `strict`-only, but wanted a second confirmation. Note that this will break compatibility with servers/clients that add some stuff to the response/request objects (one example of this is linked in the description).

jez:
> I like strict more... jsonrpc provides a way to send arbitrary data in a params field, so I don't see a reason why someone would need to use other fields

While there is `params` for requests, for results there is not. There can only be one of `result` or `error`, plus `jsonrpc` and `id`.

The use case that I have for extra top-level fields is that in some languages it's really annoying to have to deserialized untagged unions. The easy fix is to add an extra tag, that can then be ignored by tools that don't need to look for it, but if I ever want to look at a JSON RPC response that I know I serialized, I can find that extra field, treat it as a tag, and then deserialize according to the value of that tag.

I would like to urge you all to add this back as an option. We would love to make use of it in LanguageClient-neovim. Language Server Protocol clients want to be as permissive as possible to accommodate as many language servers as possible (consider: we can't fix all the language servers, so we try to accomodate them).

A flag to optionally skip the `deny_unknown_fields` would be appreciated.

jez:
I opened a follow up issue here:

https://github.com/paritytech/jsonrpc/issues/595

PR #452, opened 2019-07-16
Allow access to raw Params object in jsonrpc-derive

### LINKED ISSUE #389: derive: allow raw Params object as parameter
Idea from https://github.com/paritytech/jsonrpc/issues/387#issuecomment-463104381

Allow a single raw `Params` object, skipping the automatic deserialization in the derive macro. e.g.
```
#[rpc(name = "double", rawParams)]
fn double(&self, params: Params) -> Result<u8> {
   let addArgs: AddArgs = params.parse()?;
}
```
> It always have to be single-parameter method (variable name does not matter) and then you can parse it either as array, object or both if you need it

ebkalderon:
### Added

* Add new meta word `raw_params` to the `#[rpc]` attribute on trait methods. This is supported on both RPC and pub/sub methods.

### Changed

* Skip deserialization of `Params` if `raw_params` is set and pass it in as the sole argument instead.
* Update `meta-macros` example to include `raw_params` usage.

Closes #389.

---

The code for `jsonrpc-derive` was admittedly quite hairy, but I like how both it and the `Params` type rely quite heavily on the type system to ensure consistency. I've checked quite a few corner cases, and everything seems to be alright (e.g. rejecting multiple arguments when `raw_params` is set, rejecting non-`Params` arguments when `raw_params` is set, ensuring that pub/sub and custom metadata still work as expected with `raw_params`).

I chose the name `raw_params` over `rawParams`, as described in the issue linked above, to better adhere to idiomatic Rust style.

Please let me know if there is anything I should improve or fix further!

### DIFF
--- a/derive/examples/meta-macros.rs
+++ b/derive/examples/meta-macros.rs
@@ -1,6 +1,7 @@
 use std::collections::BTreeMap;
 
 use jsonrpc_core::futures::future::FutureResult;
+use jsonrpc_core::types::params::Params;
 use jsonrpc_core::{futures, Error, MetaIoHandler, Metadata, Result, Value};
 use jsonrpc_derive::rpc;
 
@@ -16,7 +17,7 @@ pub trait Rpc<One> {
 	#[rpc(name = "getOne")]
 	fn one(&self) -> Result<One>;
 
-	/// Adds two numbers and returns a result
+	/// Adds two numbers and returns a result.
 	#[rpc(name = "add")]
 	fn add(&self, a: u64, b: u64) -> Result<u64>;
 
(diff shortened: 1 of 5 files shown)


ebkalderon:
[clabot:check]

ebkalderon:
AppVeyor build seems to have failed prematurely due to a [connection reset error](https://ci.appveyor.com/project/paritytech/jsonrpc/builds/26005130#L514) while running tests. Travis builds appear so far to be chugging along as expected.

ebkalderon:
Thanks for the very prompt review, everyone!

PR #454, opened 2019-07-17
Add support for notifications in jsonrpc-derive

### LINKED ISSUE #252: Add notification with macro
Hello, 

I use `build_rpc_trait` macro to construct a server.
I would like add a notification (i.e. method without `id`).

As far as I can see only `add_method` is used within the macro.
Maybe we could add support for notifications while working on #66?

ebkalderon:
### Added

* Add support for notifications to the `#[rpc]` macro in `jsonrpc-derive`. Annotating a trait method that returns `()` with `#[rpc]` will register a JSON-RPC notification.

### Changed

* Add missing `IoDelegate::add_notification_with_meta()` method. This was necessary for getting the `#[notification]` attribute to support the `meta` meta word.
* Update `meta-macros` example to include notification usage.
* Update `macros` and `pubsub-macros` integration tests.

### Fixed

* Improve type inference for `raw_params` in `to_delegate.rs` (introduced with #452) with a type annotation so it works for notification methods (the only trait methods which don't return a `T: IntoFuture`).

Closes #252.

---

Note that I had to make a change to `IoDelegate` in `jsonrpc-core` to make this work. It already had `add_method()`, `add_method_with_meta()`, and `add_notification()` methods, but lacked an `add_notification_with_meta()`. I added this method so notifications could reach feature parity with standard JSON-RPC methods and pubsub methods and support metadata as well.

I've updated the `meta-macros` example to include a notification, and all seems to work well. Please let me know if there is anything I should fix or improve!

### DIFF
--- a/derive/examples/meta-macros.rs
+++ b/derive/examples/meta-macros.rs
@@ -38,7 +38,7 @@ pub trait Rpc<One> {
 	fn call_meta(&self, a: Self::Metadata, b: BTreeMap<String, Value>) -> FutureResult<String, Error>;
 
 	/// Handles notification.
-	#[notification(name = "notify")]
+	#[rpc(name = "notify")]
 	fn notify(&self, a: u64);
 }
 
--- a/derive/examples/meta-macros.rs
+++ b/derive/examples/meta-macros.rs
@@ -37,7 +37,7 @@ pub trait Rpc<One> {
 	#[rpc(meta, name = "callAsyncMeta", alias("callAsyncMetaAlias"))]
 	fn call_meta(&self, a: Self::Metadata, b: BTreeMap<String, Value>) -> FutureResult<String, Error>;
 
-	/// Responds to a notification.
+	/// Handles notification.
 	#[notification(name = "notify")]
 	fn notify(&self, a: u64);
 }
--- a/derive/examples/meta-macros.rs
+++ b/derive/examples/meta-macros.rs
@@ -36,6 +36,10 @@ pub trait Rpc<One> {
 	/// Performs an asynchronous operation with meta.
 	#[rpc(meta, name = "callAsyncMeta", alias("callAsyncMetaAlias"))]
 	fn call_meta(&self, a: Self::Metadata, b: BTreeMap<String, Value>) -> FutureResult<String, Error>;
+
+	/// Responds to a notification.
+	#[notification(name = "notify")]
+	fn notify(&self, a: u64);
 }
 
 struct RpcImpl;
@@ -64,6 +68,10 @@ impl Rpc<u64> for RpcImpl {
 
 	fn call_meta(&self, meta: Self::Metadata, map: BTreeMap<String, Value>) -> FutureResult<String, Error> {
 		futures::finished(format!("From: {}, got: {:?}", meta.0, map))
+	}
+
+	fn notify(&self, a: u64) {
+		println!("Received `notify` with value: {}", a);
 	}
 }
 
--- a/derive/src/rpc_attr.rs
+++ b/derive/src/rpc_attr.rs
@@ -41,7 +41,6 @@ const RAW_PARAMS_META_WORD: &str = "raw_params";
 const SUBSCRIBE_META_WORD: &str = "subscribe";
 const UNSUBSCRIBE_META_WORD: &str = "unsubscribe";
 const RETURNS_META_WORD: &str = "returns";
-const NOTIFICATION_ATTR_NAME: &str = "notification";
 
 const MULTIPLE_RPC_ATTRIBUTES_ERR: &str = "Expected only a single rpc attribute per method";
 const INVALID_ATTR_PARAM_NAMES_ERR: &str = "Invalid attribute parameter(s):";
@@ -187,11 +186,6 @@ fn validate_attribute_meta(meta: syn::Meta) -> Result<syn::Meta> {
 				&[SUBSCRIBE_META_WORD, UNSUBSCRIBE_META_WORD, RAW_PARAMS_META_WORD],
 			)?;
 			validate_idents(&meta, &visitor.name_value_names, &[SUBSCRIPTION_NAME_KEY, RPC_NAME_KEY])?;
-			validate_idents(&meta, &visitor.meta_list_names, &[ALIASES_KEY])
-		}
-		NOTIFICATION_ATTR_NAME => {
-			validate_idents(&meta, &visitor.meta_words, &[METADATA_META_WORD, RAW_PARAMS_META_WORD])?;
-			validate_idents(&meta, &visitor.name_value_names, &[RPC_NAME_KEY])?;
 			validate_idents(&meta, &visitor.meta_list_names, &[ALIASES_KEY])
 		}
 		_ => Ok(meta), // ignore other attributes - compiler will catch unknown ones
--- a/derive/src/rpc_attr.rs
+++ b/derive/src/rpc_attr.rs
@@ -22,6 +22,9 @@ pub enum AttributeKind {
 		subscription_name: String,
 		kind: PubSubMethodKind,
 	},
+	Notification {
+		has_metadata: bool,
+	},
 }
 
 #[derive(Clone, Debug)]
@@ -40,6 +43,7 @@ const RAW_PARAMS_META_WORD: &str = "raw_params";
 const SUBSCRIBE_META_WORD: &str = "subscribe";
 const UNSUBSCRIBE_META_WORD: &str = "unsubscribe";
 const RETURNS_META_WORD: &str = "returns";
+const NOTIFICATION_ATTR_NAME: &str = "notification";
 
 const MULTIPLE_RPC_ATTRIBUTES_ERR: &str = "Expected only a single rpc attribute per method";
 const INVALID_ATTR_PARAM_NAMES_ERR: &str = "Invalid attribute parameter(s):";
@@ -74,6 +78,11 @@ impl RpcMethodAttribute {
 						Some(Ok(AttributeKind::Rpc { has_metadata, returns }))
 					}
 					PUB_SUB_ATTR_NAME => Some(Self::parse_pubsub(meta)),
+					NOTIFICATION_ATTR_NAME => {
+						let has_metadata =
+							get_meta_list(meta).map_or(false, |ml| has_meta_word(METADATA_META_WORD, ml));
+						Some(Ok(AttributeKind::Notification { has_metadata }))
+					}
 					_ => None,
 				};
 				attr_kind.map(|kind| {
@@ -122,7 +131,14 @@ impl RpcMethodAttribute {
 	pub fn is_pubsub(&self) -> bool {
 		match self.kind {
 			AttributeKind::PubSub { .. } => true,
-			AttributeKind::Rpc { .. } => false,
+			_ => false,
+		}
+	}
+
+	pub fn is_notification(&self) -> bool {
+		match self.kind {
+			AttributeKind::Notification { .. } => true,
+			_ => false,
 		}
 	}
 }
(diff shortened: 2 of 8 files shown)


ebkalderon:
This `trybuild` [linking error in Travis CI](https://travis-ci.org/paritytech/jsonrpc/jobs/559891824#L292-L308) also occurred on my machine and I solved it by running `cargo clean` to clear my build cache before running `cargo test` again. Looks like something must have broken upstream in the `trybuild` crate recently, as it was [last updated 18 hours ago](https://crates.io/crates/trybuild), at the time of writing.

derive/src/to_client.rs
```diff
@@ -134,6 +132,7 @@ fn generate_client_methods(methods: &[MethodRegistration]) -> Result<Vec<syn::Im
 				);
 				client_methods.push(client_method);
 			}
+			_ => continue,
```
ebkalderon: I put this fallback case here because I am still waiting on an answer for https://github.com/paritytech/jsonrpc/issues/252#issuecomment-512122117 and am currently not handling the `MethodRegistration::Notification` case in the client code generation.

derive/examples/meta-macros.rs
```diff
@@ -36,6 +36,10 @@ pub trait Rpc<One> {
 	/// Performs an asynchronous operation with meta.
 	#[rpc(meta, name = "callAsyncMeta", alias("callAsyncMetaAlias"))]
 	fn call_meta(&self, a: Self::Metadata, b: BTreeMap<String, Value>) -> FutureResult<String, Error>;
+
+	/// Responds to a notification.
+	#[notification(name = "notify")]
```
tomusdrw: Do we actually need a different attribute? Can't we detect that the return type is `()` and act upon that?ac

derive/examples/meta-macros.rs
```diff
@@ -36,6 +36,10 @@ pub trait Rpc<One> {
 	/// Performs an asynchronous operation with meta.
 	#[rpc(meta, name = "callAsyncMeta", alias("callAsyncMetaAlias"))]
 	fn call_meta(&self, a: Self::Metadata, b: BTreeMap<String, Value>) -> FutureResult<String, Error>;
+
+	/// Responds to a notification.
```
tomusdrw:
```suggestion
	/// Handles notification.
```
since there is no response really.

derive/src/to_client.rs
```diff
@@ -97,9 +97,7 @@ fn generate_client_methods(methods: &[MethodRegistration]) -> Result<Vec<syn::Im
 				let arg_names = compute_arg_identifiers(&args)?;
 				let returns = match &method.attr.kind {
 					AttributeKind::Rpc { returns, .. } => compute_returns(&method.trait_item, returns)?,
-					AttributeKind::PubSub { .. } => {
-						continue;
-					}
+					_ => continue,
```
tomusdrw: I'd prefer to list the variants here, since modifying `AttributeKind` in the future will force handling the case here.

derive/src/to_client.rs
tomusdrw: Please rather put the variant instead of a wildcard.

derive/src/to_delegate.rs
```diff
@@ -199,7 +218,7 @@ impl RpcMethod {
 			} else if param_types.is_empty() {
 				quote! { let params = params.expect_no_params(); }
 			} else if self.attr.raw_params {
-				quote! { let params = Ok((params,)); }
+				quote! { let params: _jsonrpc_core::Result<_> = Ok((params,)); }
```
tomusdrw:
```suggestion
				quote! { let params = _jsonrpc_core::Result::Ok((params,)); }
```

derive/tests/macros.rs
```diff
@@ -14,20 +14,25 @@ type Result<T> = ::std::result::Result<T, MyError>;
...
-	/// Adds two numbers and returns a result
+	/// Adds two numbers and returns a result.
 	#[rpc(name = "add", alias("add_alias1", "add_alias2"))]
 	fn add(&self, a: u64, b: u64) -> Result<u64>;
 
+	/// Retrieves and debug prints the underlying `Params` object.
 	#[rpc(name = "raw", raw_params)]
 	fn raw(&self, params: Params) -> Result<String>;
+
+	/// Responds to a notification.
```
tomusdrw:
```suggestion
	/// Handles a notification.
```

derive/examples/meta-macros.rs
ebkalderon: I think that works. We can certainly do that!

ebkalderon: The main reason why I chose a different attribute is so we can reject the `returns` key from being used with notifications, but I think we could probably make that work. Maybe we could emit an error for functions annotated with `returns` that also return a `()`?

derive/src/to_client.rs
ebkalderon: Sounds good. But what do we do about the client code generation for notification support? Is there anything we need to add to the `MethodRegistration::Notification` case?

derive/src/to_client.rs
ebkalderon: Will do, once I have an answer for https://github.com/paritytech/jsonrpc/pull/454#discussion_r304851662.

derive/src/to_delegate.rs
ebkalderon: This sadly won't compile. We need a type annotation here for this to work in all cases.

derive/examples/meta-macros.rs
ebkalderon: Okay, I believe this is doable! We could remove the `AttributeKind::Notification` variant and add an `is_notification: bool` field to `AttributeKind::Rpc`.

This is necessary because without some kind of marker as to whether an `rpc` method is a true method or a notification, we would need to check the return type twice, both in `rpc_trait.rs` and `to_delegate.rs`. Doing so just once in `parse_meta()` inside `rpc_attr.rs` when we first determine the `AttributeKind` is the most efficient way to do this, I think.

We can also use this opportunity to make sure that the `returns` meta word isn't being used on notifications.

ebkalderon: Okay, I believe your suggestion is doable! We could remove the `AttributeKind::Notification` variant and add an `is_notification: bool` field to `AttributeKind::Rpc`.

This is necessary because without some kind of marker as to whether an `rpc` method is a true method or a notification, we would need to check the return type twice, both in `rpc_trait.rs` and `to_delegate.rs`. Doing so just once in `parse_meta()` inside `rpc_attr.rs` when we first determine the `AttributeKind` is the most efficient way to do this, I think.

We can also use this opportunity to make sure that the `returns` meta word isn't being used on notifications.

derive/src/to_client.rs
ebkalderon: I've put the explicit value here, and I'm currently not generating anything for notifications in the client code. If we need support for this, please respond to https://github.com/paritytech/jsonrpc/pull/454#discussion_r304851662 and I can add it to this PR.

derive/examples/meta-macros.rs
ebkalderon: Done in https://github.com/paritytech/jsonrpc/pull/454/commits/f4f20fc07f5c451f2c956c04af6fbce0b91e8e3f.

derive/src/to_client.rs
ebkalderon: I have added the explicit variant in place of the wildcard. I am also assuming we do not need to generate any client code for notifications. If we need support for this, please let me know and I can implement it in this PR.

ebkalderon:
AppVeyor build seems to have failed prematurely due to a [connection reset error](https://ci.appveyor.com/project/paritytech/jsonrpc/builds/26123509) while running tests. Travis builds appear so far to be chugging along as expected.

ebkalderon:
@tomusdrw @ascjones Looks like the linking issue with the `trybuild` crate is occurring again on Travis CI macOS builds. The PR itself is ready for review!

derive/src/rpc_attr.rs
```diff
@@ -40,6 +41,7 @@ const RAW_PARAMS_META_WORD: &str = "raw_params";
 const SUBSCRIBE_META_WORD: &str = "subscribe";
 const UNSUBSCRIBE_META_WORD: &str = "unsubscribe";
 const RETURNS_META_WORD: &str = "returns";
+const NOTIFICATION_ATTR_NAME: &str = "notification";
```
tomusdrw: Is this still in use? Seems not needed in the example.

derive/src/to_delegate.rs
tomusdrw: Oh, yeah. You're right. Sorry!

derive/tests/pubsub-macros.rs
```diff
@@ -22,17 +22,21 @@ type Result<T> = ::std::result::Result<T, MyError>;
...
 	#[pubsub(subscription = "hello", unsubscribe, name = "hello_unsubscribe")]
 	fn unsubscribe(&self, a: Option<Self::Metadata>, b: SubscriptionId) -> Result<bool>;
 
-	/// A regular rpc method alongside pubsub
+	/// A regular rpc method alongside pubsub.
 	#[rpc(name = "add")]
 	fn add(&self, a: u64, b: u64) -> Result<u64>;
+
+	/// A notification alongside pubsub.
+	#[rpc(name = "notify")]
```
tomusdrw: Can we get a test for notifications to make sure that it actually only accepts a request with no id for that method?

derive/src/rpc_attr.rs
```diff
@@ -162,6 +189,11 @@ fn validate_attribute_meta(meta: syn::Meta) -> Result<syn::Meta> {
 			validate_idents(&meta, &visitor.name_value_names, &[SUBSCRIPTION_NAME_KEY, RPC_NAME_KEY])?;
 			validate_idents(&meta, &visitor.meta_list_names, &[ALIASES_KEY])
 		}
+		NOTIFICATION_ATTR_NAME => {
```
tomusdrw: Same here, I don't think we use it any more, since we use the output type now.

derive/tests/pubsub-macros.rs
ebkalderon: Sure! I'll add a test for that.

derive/src/rpc_attr.rs
ebkalderon: Whoops, I forgot to remove the attribute from the code while making the changes. Good catch! 👍

derive/src/rpc_attr.rs
ebkalderon: Good catch! I'll remove the unused attribute from the code right away.

derive/tests/pubsub-macros.rs
ebkalderon: Okay, I've added a new test case `should_accept_only_notifications()` which should cover this.

ebkalderon:
Thank you for the suggestions, @tomusdrw! I've removed the unused `NOTIFICATION_ATTR_NAME` constant and added a new test case which verifies that the `#[rpc]` macro is correctly registering a notification.

derive/src/to_client.rs
```diff
@@ -134,6 +132,7 @@ fn generate_client_methods(methods: &[MethodRegistration]) -> Result<Vec<syn::Im
 				);
 				client_methods.push(client_method);
 			}
+			MethodRegistration::Notification { .. } => continue,
```
ascjones: Users may wonder why client notification methods are not generated. Not sure of the best approach here - we used to [print a message during compilation](https://github.com/paritytech/jsonrpc/pull/436/files#diff-77384f93cc36bb6db93182b47a771715L115) for pubsub but that was annoying. Alternatively could generate the method with `unimplemented!()`?

Also would be good to have an issue "Implement client support for notifications"

tomusdrw: I like the `unimplemented!()` idea, if that's not too much effort to get in.

ascjones: https://github.com/paritytech/jsonrpc/issues/461

ebkalderon: Excellent idea! I'll add that in.

ebkalderon:
@ascjones @tomusdrw Done! I've added a dummy client method for notifications that calls `unimplemented!()`.

PR #455, opened 2019-07-17
Add client transport over unix domain socket

jkryl:
I have noticed there is IPC server but no client implementation. I needed that for my project so I decided to add it. It's not IPC but unix domain socket (UDS) transport. So it does not work on windows unlike IPC, but I hope it can be still useful for people until it is replaced by IPC client.

A couple of remarks about the fix itself. Without much knowledge of jsonrpc implementation I took websocket transport implementation as a starting point and made minimal changes to it to make it work for UDS. I added three new test cases. All I ever tested are single method calls. I did not test subscribe functionality as I don't need that in my project. I have chosen newline as the delimiter between the messages sent over UDS. The uds feature is optional and turned off by default.

I will appreciate any help with making the PR to meet quality standards of jsonrpc code. Thanks!

### DIFF
--- a/core-client/transports/src/transports/uds.rs
+++ b/core-client/transports/src/transports/uds.rs
@@ -1,8 +1,7 @@
 //! JSON-RPC unix domain socket client implementation.
 use crate::{RpcChannel, RpcError};
-use failure::Error;
 use futures::prelude::*;
-use std::{collections::VecDeque, path::Path};
+use std::path::Path;
 use tokio::codec::{Framed, LinesCodec};
 use tokio_uds::UnixStream;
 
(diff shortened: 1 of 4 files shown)


jkryl:
[clabot:check]

ebkalderon:
@jkryl Looks like your Travis CI build [failed for the same reason as my PR](https://github.com/paritytech/jsonrpc/pull/454#issuecomment-512201193). It seems that `trybuild`, an upstream dependency, was updated recently and has some linking issues on Linux. I solved the problem locally by running `cargo clean`, but I imagine the project's Travis `sccache` dir is now poisoned.

CC @ascjones @tomusdrw.

jkryl:
thanks @ebkalderon for the notice. I fixed the formatting of the code, pushed again and now it succeeded. Most likely someone has fixed the problem in the meantime.

core-client/transports/src/transports/uds.rs
```diff
@@ -0,0 +1,178 @@
...
+	pub fn new(sink: TSink, stream: TStream) -> Self {
+		Self {
+			sink,
+			stream,
+			queue: VecDeque::new(),
+		}
+	}
+}
+
+impl<TSink, TStream, TError> Sink for UdsClient<TSink, TStream>
```
ascjones: This is essentially the same to the ws `Sink` impl. https://github.com/paritytech/jsonrpc/blob/master/core-client/transports/src/transports/ws.rs#L51. I wonder whether, if this is required for other duplex transports, we should abstract it.

core-client/transports/src/transports/uds.rs
```diff
@@ -0,0 +1,178 @@
...
+					Err(error) => return Err(RpcError::Other(error.into())),
+				},
+				None => break,
+			}
+		}
+		self.sink.poll_complete().map_err(|error| RpcError::Other(error.into()))
+	}
+}
+
+impl<TSink, TStream, TError> Stream for UdsClient<TSink, TStream>
```
ascjones: Could this be replaced with calling `Stream::map_err` in `connect`?

jkryl: I will experiment with it next week. Would be great if we could avoid some of this code. thanks!

Xanewok:
As it turns out I also needed that for a project, albeit in a cross-platform solution (including Windows). I tried to use the `parity_tokio_ipc::IpcConnection` which resulted in:

```rust

use futures::sink::Sink;
use futures::stream::Stream;
use futures::Future;
use jsonrpc_core::types::params::Params;
use jsonrpc_core_client::transports::duplex;
use jsonrpc_core_client::RpcError;
use jsonrpc_server_utils::codecs::StreamCodec;
use parity_tokio_ipc::IpcConnection;
use tokio::codec::Decoder;
use tokio::runtime::Runtime;

// ...
let mut runtime = Runtime::new().unwrap();
// Handle::default() only doesn't work specifically on Windows, not sure yet why
let handle = runtime.reactor();

let connection = IpcConnection::connect(path, handle)?;
let codec = jsonrpc_server_utils::codecs::StreamCodec::stream_incoming();

let (sink, stream) = codec.framed(connection).split();
let sink = sink.sink_map_err(|e| RpcError::Other(e.into()));
let stream = stream.map_err(|e| RpcError::Other(e.into()));

let (client, sender) = duplex(sink, stream);

let raw = jsonrpc_core_client::RawClient::from(sender);
let result = raw
    .call_method("say_hello", Params::None)
    .map(|val| eprintln!("ipc: say_hello: {:?}", val))
    .map_err(|e| eprintln!("ipc: say_hello failed: {:?}", e));
let client = client.map_err(|e| panic!("Client err: {:?}", e));

runtime.spawn(client);

dbg!(&runtime.block_on(result));
```

and it works from what I can tell. Am I missing something? I understand this lacks message buffering like in the other Client implementations?

jkryl:
I used map error functions as suggested by @ascjones and got basically what @Xanework has. @Xanewok: could you submit a PR with your changes? IPC is preferable over UDS? Perhaps you could add the 3 test cases slightly modified for IPC if you have none. If you agree with the plan I will close this PR.

Xanewok:
@jkryl submitted https://github.com/paritytech/jsonrpc/pull/460 with the IPC implementation and your tests adapted. I'd be grateful if you could also review it :heart:

jkryl:
I'm closing this PR in favour of #460 .

@Xanewok : I will test your ipc client in my project this week and see how it works. I will update your PR then. Thanks for the fix.

PR #459, opened 2019-07-20
ipc: Allow specifying different event loop reactor in server builder

Xanewok:
Right now we use `Handle::default()` (which should lazily bind to a reactor) when creating an `IpcConnection` but that surprisingly doesn't work with either the user-specified executor or the lazily spawned pool as the fallback executor on Windows (only tested on Windows 10).

Using a simple [`Runtime::reactor`](https://docs.rs/tokio/0.1.22/tokio/runtime/struct.Runtime.html#method.reactor) works, so to work around the issue I'd like be able to specify a custom reactor to be used (which seems somewhat useful on its own, hopefully).

Disclaimer: I'm not yet well versed in the async-fu so I might be missing something obvious here, please correct me if I'm wrong.

### DIFF
--- a/ipc/src/server.rs
+++ b/ipc/src/server.rs
@@ -50,7 +50,7 @@ pub struct ServerBuilder<M: Metadata = (), S: Middleware<M> = middleware::Noop> 
 	meta_extractor: Arc<dyn MetaExtractor<M>>,
 	session_stats: Option<Arc<dyn session::SessionStats>>,
 	executor: reactor::UninitializedExecutor,
-	reactor: Handle,
+	reactor: Option<Handle>,
 	incoming_separator: codecs::Separator,
 	outgoing_separator: codecs::Separator,
 	security_attributes: SecurityAttributes,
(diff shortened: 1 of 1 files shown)


Xanewok:
[clabot:check]

Xanewok:
Yeah, it seems that named pipes require a consistent thread for the wakeups (cc https://github.com/NikVolf/tokio-named-pipes/pull/2 and @c0gent who authored the change) so it seems necessary to be able to provide a user-specified handle.

Not sure if this changes semantics as we now construct `Handle::default()` outside the lazily spawned future now (https://github.com/tokio-rs/tokio/issues/307)
> The difference is that `Handle::default` will use the global reactor if called from outside of a runtime. This global reactor will most likely go away in 0.2 and was from pre Runtime.

I can use `Option<Handle>` for `ServerBuilder::reactor` and then inside the future use
```rust
let reactor = self.reactor.unwrap_or_else(Handle::default);
```
to bring back the original behaviour in case user didn't override the previous default reactor.

tomusdrw:
> can use Option<Handle> for ServerBuilder::reactor and then inside the future use..

@Xanewok yeah, that makes sense to me

Xanewok:
Done, added a comment why we need to create a default value inside a future

Xanewok:
```
test transports::http::tests::handles_connection_refused_error ... test transports::http::tests::handles_connection_refused_error has been running for over 60 seconds
```

that's unfortunate

PR #460, opened 2019-07-21
Add an IPC client transport using parity-tokio-ipc

Xanewok:
An IPC version of https://github.com/paritytech/jsonrpc/pull/455.

Key differences:
- `parity_tokio_ipc::IpcConnection` requires an already running event loop (can probably be adapted upstream to use futures-only style) whereas `UnixStream::connect` (which it wraps on Unices) returns only a future
- Uses a stream codec (from server-utils) rather than newline-separated messages

As I mentioned in https://github.com/paritytech/jsonrpc/pull/459 I encountered a bug where the server will not be polled using lazily-bound I/O reactor. To work around it, in the test cases I need to signal it to stop using `server.close()` to make it progress, which is less than ideal (but makes it work cross-platform). Being able to specify custom reactor in ServerBuilder (https://github.com/paritytech/jsonrpc/pull/459) fixes the issue.

I'm not sure whether it's because of how `IpcConnection` is created. Maybe moving it to future-only style would fix the reactor binding on Windows? I'll experiment and get back with results.

### DIFF
--- a/core-client/transports/src/transports/ipc.rs
+++ b/core-client/transports/src/transports/ipc.rs
@@ -1,0 +1,125 @@
+//! JSON-RPC IPC client implementation using Unix Domain Sockets on UNIX-likes
+//! and Named Pipes on Windows.
+
+use crate::transports::duplex::duplex;
+use crate::{RpcChannel, RpcError};
+use futures::prelude::*;
+use jsonrpc_server_utils::codecs::StreamCodec;
+use parity_tokio_ipc::IpcConnection;
+use std::io;
+use std::path::Path;
+use tokio::codec::Decoder;
+
+/// Connect to a JSON-RPC IPC server.
+pub fn connect<P: AsRef<Path>, Client: From<RpcChannel>>(
+	path: P,
+	reactor: &tokio::reactor::Handle,
+) -> Result<impl Future<Item = Client, Error = RpcError>, io::Error> {
+	let connection = IpcConnection::connect(path, reactor)?;
+
+	Ok(futures::lazy(move || {
+		let (sink, stream) = StreamCodec::stream_incoming().framed(connection).split();
+		let sink = sink.sink_map_err(|e| RpcError::Other(e.into()));
+		let stream = stream.map_err(|e| RpcError::Other(e.into()));
+
+		let (client, sender) = duplex(sink, stream);
+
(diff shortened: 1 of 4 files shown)


Xanewok:
https://ci.appveyor.com/project/paritytech/jsonrpc/builds/26132607#L511
```
test transports::http::tests::handles_connection_refused_error ...
test transports::http::tests::handles_connection_refused_error has been running for over 60 seconds
```

I don't think this should impact HTTP transport?

core-client/transports/src/transports/ipc.rs
```diff
@@ -0,0 +1,125 @@
...
+		server.close();
+
+		match rt.block_on(fut) {
+			Ok(val) => assert_eq!(&val, "Hello Jeffry!"),
+			Err(err) => panic!("IPC RPC call failed: {}", err),
+		}
+		rt.shutdown_now().wait().unwrap();
+	}
+
+	// Windows currently panics in `IpcConnection::connect` if a given path doesn't exist
```
tomusdrw: Can we add this link to the comment: https://github.com/NikVolf/parity-tokio-ipc/pull/12 so that we know we should fix it when it's in?

Xanewok: Done in https://github.com/paritytech/jsonrpc/pull/460/commits/5bb62b59cba9b2a9b12a37383e5d0fbcab318054; thankfully the PR is already merged so this now only needs a point release

ascjones:
@Xanewok just needs a conflict resolving then we can merge :+1:

Xanewok:
Done!

jkryl:
@Xanewok : the fix works for me 👍  The connect function requires reactor handle to be passed as the 2nd argument which seems to be impossible without using deprecated `rt.reactor()` (as done in the tests). Just curious if there is a way how to do this without using deprecated tokio api? I mean this warning:

```
warning: use of deprecated item 'tokio::runtime::Runtime::reactor': there is now a reactor per worker thread
  --> src/main.rs:60:21
   |
60 |     let handle = rt.reactor().clone();
   |                     ^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
```

Xanewok:
@jkryl see https://github.com/paritytech/jsonrpc/pull/459 for the Windows-specific problem, I believe `Handle::default()` works in Unices. Can you try that instead?

jkryl:
@Xanewok : thanks!

Issue #463, opened 2019-07-22
Batch requests from client


ascjones:
Support batch requests in `core-client` transports, and potentially in the `derive` client if we can come up with a nice API.

tomusdrw:
In `rust-web3` I've implemented batch as a transport wrapper (i.e. defers sending until you tell it too).

For `derive()` I think an API like this would be nice:
```
// regular call
let result = client.call_something();

// batch call
let batch = client.batch();
let result = batch.call_something();
spawn(batch.send());
```

PR #473, opened 2019-08-05
Add a method to handle empty response.

tomusdrw:
Closes: #463 

>  If there are no Response objects contained within the Response array as it is to be sent to the client, the server MUST NOT return an empty Array and should return nothing at all.

But the issue is that "empty response" is not valid JSON (at least according to serde), so it's not possible to directly decode empty string to anything.

I'm a bit torn apart by this issue, since on one hand we can just ask the clients to check the empty response, but as a compromise this PR adds a `Response::from_json` helper method to parse the string into `Response` type and handle an empty string case.

@debris please let me know what you think of it.

### DIFF
--- a/core/src/io.rs
+++ b/core/src/io.rs
@@ -530,6 +530,24 @@ mod tests {
 	}
 
 	#[test]
+	fn test_batch_notification() {
+		use std::sync::atomic;
+		use std::sync::Arc;
+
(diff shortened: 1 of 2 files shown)


leoyvens:
We once had a bug because of this, thanks for fixing!

PR #479, opened 2019-08-19
[client] don't block other subscriptions if subscriber not ready

ascjones:
In [`subxt`](https://github.com/paritytech/substrate-subxt/blob/master/src/rpc.rs#L310) we start a subscription for `events` prior to submitting an extrinsic, and only read from it once the extrinsic has been finalized, in order to extract the resulting events.

However, because we are not reading from the `events` stream, the `Sender` sink returns `Async::NotReady` which currently results in the message being pushed back onto the front of the `incoming` queue. This will block messages for the `extrinsic` subscription, and so we are deadlocked: we will not read `events` until `extrinsics` is successful, and `extrinsics` messages are blocked.

This PR pushes a message for a subscriber which is not ready to the back of the queue, in order not to block other incoming messages.

### DIFF
--- a/core-client/transports/src/transports/duplex.rs
+++ b/core-client/transports/src/transports/duplex.rs
@@ -247,7 +247,7 @@ where
 							}
 							Ok(Async::NotReady) => {
 								let (sid, method) = sid_and_method;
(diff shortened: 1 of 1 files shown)


tomusdrw:
Sounds reasonable.

PR #480, opened 2019-08-20
Backport #479, bump core-client patch version

ascjones:


ascjones:
Unfortunately not, tried and failed to write one.

PR #483, opened 2019-08-26
Make IoHandlers cloneable.

tomusdrw:
CC @tomaka


dvdplm:
What is the rationale for this?

tomusdrw:
@dvdplm There are no reasons why they wouldn't be cloneable. And Pierre bumped into this when refactoring substrate code.

PR #488, opened 2019-09-12
Add `iter` and `IntoIterator` to `IoHandler`

tomusdrw:
Context: To be able to get all methods current in `IoHandler` and add a "meta" call that returns all available methods.


tomusdrw:
@dvdplm `Iter` and `IntoIter` sound very generic, I think I'd be most inclined to import `hash_map` module and write `hash_map::Iter` and `hash_map::IntoIter` to at least see where the type is coming from, but I'm not super strong on this, so fixed it.

dvdplm:
> import `hash_map` module and write `hash_map::Iter` and `hash_map::IntoIter`

Yeah, probably better still.

Issue #481, opened 2019-08-22
#[rpc] macro in derive has dependencies which are not propagated into consuming projects


fwcd:
I am trying to implement the [basic example using derive](https://github.com/paritytech/jsonrpc#basic-usage-with-derive) from the README. Since the code snippet only seems to require `jsonrpc-core` and `jsonrpc-derive`, I did not include any other dependencies in `Cargo.toml`:

```toml
[dependencies]
jsonrpc-core = "13.0.0"
jsonrpc-derive = "13.0.0"
```

The compiler, however, yields an error that a crate named `serde` is required too:

```
error: Could not find `serde` in `dependencies` or `dev-dependencies` in `...\Cargo.toml`!
 --> src\main.rs:4:1
  |
4 | #[rpc]
  | ^^^^^^
```

Indeed does the `#[rpc]` macro have a dependency on `serde`:

https://github.com/paritytech/jsonrpc/blob/54c4de527f536f8b6c5c6f4616446f82d00ed936/derive/src/rpc_trait.rs#L258-L261

`serde` is only listed as a dev-dependency though, so it [will not propagate](https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html) into projects consuming the library:

https://github.com/paritytech/jsonrpc/blob/54c4de527f536f8b6c5c6f4616446f82d00ed936/derive/Cargo.toml#L21-L30

Adding `serde` explicitly to my Cargo.toml did fix the issue, but brought up another missing dependency (`jsonrpc-core-client`) which, again, had to be added manually despite being used by the `#[rpc]` macro.

Am I missing something? Would it be possible to clarify which dependencies have to be provided manually when using `jsonrpc-derive`?

dvdplm:
This sounds like a bug to me. Want to attempt a PR to fix it? :)

tomusdrw:
Yeah, so this is a tricky one. The dependencies can't be just in `jsonrpc-derive` crate - we would need to re-export them, because the macro generates code in the crate when it's used.

However the set of crates that is needed depends on the features you're going to use. e.g.` jsonrpc-core-client` is only necessary if the client code is being generated.

I'm ok with re-exporting deps that are 100% required (like `serde`), but have mixed feelings about `jsonrpc-core-client` for instance.

simeir:
> Yeah, so this is a tricky one. The dependencies can't be just in `jsonrpc-derive` crate - we would need to re-export them, because the macro generates code in the crate when it's used.
[...]

PR #498, opened 2019-10-09
Solve compiler error when serde is not a dependency of user project (#481)

simeir:
Use `serde` directly instead of ask for user of the library to add it as a
dependency.

Tests are modified since this is a ui change.

The doc comment gives an example using `#[rpc(server)]` instead of original
`#[rpc]`, telling users that this attribute has an option to be used to
configure it.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -30,6 +30,8 @@ pub use futures;
 
 #[doc(hidden)]
 pub extern crate serde_json;
+#[doc(hidden)]
+pub extern crate serde;
 
 mod calls;
 mod io;
--- a/derive/src/to_delegate.rs
+++ b/derive/src/to_delegate.rs
@@ -449,17 +449,17 @@ pub fn generate_where_clause_serialization_predicates(
 			// add json serialization trait bounds
 			if client {
 				if visitor.server_to_client_type_params.contains(&ty.ident) {
-					bounds.push(parse_quote!(_serde::de::DeserializeOwned))
+					bounds.push(parse_quote!(_jsonrpc_core::serde::de::DeserializeOwned))
 				}
 				if visitor.client_to_server_type_params.contains(&ty.ident) {
-					bounds.push(parse_quote!(_serde::Serialize))
+					bounds.push(parse_quote!(_jsonrpc_core::serde::Serialize))
 				}
 			} else {
 				if visitor.server_to_client_type_params.contains(&ty.ident) {
-					bounds.push(parse_quote!(_serde::Serialize))
+					bounds.push(parse_quote!(_jsonrpc_core::serde::Serialize))
 				}
 				if visitor.client_to_server_type_params.contains(&ty.ident) {
-					bounds.push(parse_quote!(_serde::de::DeserializeOwned))
+					bounds.push(parse_quote!(_jsonrpc_core::serde::de::DeserializeOwned))
 				}
 			}
 			syn::WherePredicate::Type(syn::PredicateType {
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -9,7 +9,7 @@
 //! use jsonrpc_core::{IoHandler, Error, Result};
 //! use jsonrpc_core::futures::future::{self, FutureResult};
 //!
-//! #[rpc]
+//! #[rpc(server)]
 //! pub trait Rpc {
 //! 	#[rpc(name = "protocolVersion")]
 //! 	fn protocol_version(&self) -> Result<String>;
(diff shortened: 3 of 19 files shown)


simeir:
> This actually happens to be even worse than previously, cause previously we were allowing to rename `serde` in the `Cargo.toml` (see `crate_name("serde")`) now it has to be named `serde` and still must be imported in the top level `Cargo.toml`.
> 
> We should rather re-export `serde` from `jsonrpc-core` or `derive` and use that instead of relying on `serde` directly.

I actually tested using another crate and this does not seem to be case. I can check again, though.

simeir:
[clabot:check]

derive/src/to_delegate.rs
```diff
@@ -449,17 +449,17 @@ pub fn generate_where_clause_serialization_predicates(
 			// add json serialization trait bounds
 			if client {
 				if visitor.server_to_client_type_params.contains(&ty.ident) {
-					bounds.push(parse_quote!(_serde::de::DeserializeOwned))
+					bounds.push(parse_quote!(jsonrpc_core::serde::de::DeserializeOwned))
```
tomusdrw: please use `#core_name` instead of `jsonrpc_core`. This allows using:
```
rpc = { package = "jsonrpc-core", ... }
```
inside the `Cargo.toml` - so the crate is renamed, and in such case your `jsonrpc_core` import would not work.

Note that the code we generate here is running in the context of the user's crate, so we don't want to enforce them to import `jsonrpc-core` in a specific way.
```suggestion
					bounds.push(parse_quote!(#core_name::serde::de::DeserializeOwned))
```

derive/src/to_delegate.rs
```diff
@@ -449,17 +449,17 @@ pub fn generate_where_clause_serialization_predicates(
...
 				}
 				if visitor.client_to_server_type_params.contains(&ty.ident) {
-					bounds.push(parse_quote!(_serde::Serialize))
+					bounds.push(parse_quote!(jsonrpc_core::serde::Serialize))
```
tomusdrw:
```suggestion
					bounds.push(parse_quote!(#core_name::serde::Serialize))
```

derive/src/to_delegate.rs
```diff
@@ -449,17 +449,17 @@ pub fn generate_where_clause_serialization_predicates(
...
 				}
 			} else {
 				if visitor.server_to_client_type_params.contains(&ty.ident) {
-					bounds.push(parse_quote!(_serde::Serialize))
+					bounds.push(parse_quote!(jsonrpc_core::serde::Serialize))
```
tomusdrw:
```suggestion
					bounds.push(parse_quote!(#core_name::serde::Serialize))
```

derive/src/to_delegate.rs
```diff
@@ -449,17 +449,17 @@ pub fn generate_where_clause_serialization_predicates(
...
 				}
 				if visitor.client_to_server_type_params.contains(&ty.ident) {
-					bounds.push(parse_quote!(_serde::de::DeserializeOwned))
+					bounds.push(parse_quote!(jsonrpc_core::serde::de::DeserializeOwned))
```
tomusdrw:
```suggestion
					bounds.push(parse_quote!(#core_name::serde::de::DeserializeOwned))
```

tomusdrw:
Re: CLA, we require signing it off mainly to be able to change license in the future (if needed) without getting a consent from all contributors. Not a lawyer though (and this is not an advice), so please make sure you read it and understand all the aspects.

core/src/lib.rs
```diff
@@ -30,6 +30,7 @@ pub use futures;
 
 #[doc(hidden)]
 pub extern crate serde_json;
+pub extern crate serde;
```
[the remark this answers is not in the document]
tomusdrw: This is to prevent having all `serde` documentation in `jsonrpc_core` documentation.

simeir:
> Re: CLA, we require signing it off mainly to be able to change license in the future (if needed) without getting a consent from all contributors. Not a lawyer though (and this is not an advice), so please make sure you read it and understand all the aspects.

Got it. Thanks!

derive/src/to_delegate.rs
simeir: I think it might be better to have `_jsonrpc_core` instead of `#core_name`, as the name is already imported as `_jsonrpc_core` and all other parts of the code are using `_jsonrpc_core` to refer to the crate.

core/src/lib.rs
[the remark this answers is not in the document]
simeir: Added this.

PR #501, opened 2019-10-14
Fix race condition on `wait()`

foriequal0:
`wait()` using `oneshot::channel()` makes race conditions. Call `Executor::wait()` directly.

It is possible that the thread that has executed `wait()`, which is usually the main thread, terminates before the thread executing `done_tx.send(())` drops `rpc_handler`.
Static variables are destructed at the end of termination of the main
thread, and then a segfault occurs when the thread dropping the
`rpc_handler` accesses the variables.

### DIFF
--- a/http/src/lib.rs
+++ b/http/src/lib.rs
@@ -45,7 +45,7 @@ use parking_lot::Mutex;
 use crate::jsonrpc::futures::sync::oneshot;
 use crate::jsonrpc::futures::{self, Future, Stream};
 use crate::jsonrpc::MetaIoHandler;
-use crate::server_utils::reactor::{Executor, UninitializedExecutor};
+use crate::server_utils::reactor::{Executor, ExecutorCloseHandle, UninitializedExecutor};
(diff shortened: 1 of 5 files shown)


foriequal0:
[clabot:check]

foriequal0:
@tomusdrw I just realized that I can fix this issue with a few lines of changes while not changing the behavior. 
`Future::select` pass the unresolved future, which is `server` holding `rpc_handler`, to the following chain. Therefore, it is dropped after the `done_tx.send(())`. I didn't know that this is the reason at that time.
Dropping them before `send` is enough to fix this issue, and it doesn't change the behaviour.
1. Here: https://github.com/paritytech/jsonrpc/blob/master/http/src/lib.rs#L568
2. and here: https://github.com/paritytech/jsonrpc/blob/master/ipc/src/server.rs#L243

Should I revert the commit and proceed with this fix?

foriequal0:
I'll close this and create a new PR(#504) with the suggested fix

PR #502, opened 2019-10-15
ignore dropped connections

### LINKED ISSUE #500: Ignore connections that have no peer_addr
If there's no `peer_addr` most likely the connection has been dropped, I think in this case we should ignore the socket.

https://github.com/paritytech/jsonrpc/blob/44184766aeb59a95b4d6f4c7eaff0c9e1ef278eb/tcp/src/server.rs#L87-L93

seunlanlege:
Previously If there's no `peer_addr` on the socket, this causes the tcp-server to panic. 

https://github.com/paritytech/jsonrpc/blob/44184766aeb59a95b4d6f4c7eaff0c9e1ef278eb/tcp/src/server.rs#L87-L93

Instead of panicking this PR instead ignores the connection as it has most likely been dropped by the client.

closes #500

### DIFF
--- a/tcp/src/server.rs
+++ b/tcp/src/server.rs
@@ -91,7 +91,7 @@ impl<M: Metadata, S: Middleware<M> + 'static> ServerBuilder<M, S> {
 					let peer_addr = match socket.peer_addr() {
 						Ok(addr) => addr,
 						Err(e) => {
-							trace!(target: "tcp", "Unable to determine socket peer address, ignoring connection {}", e);
+							warn!(target: "tcp", "Unable to determine socket peer address, ignoring connection {}", e);
 							return future::Either::A(future::ok(()))
 						}
 					};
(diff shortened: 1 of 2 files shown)


dvdplm:
PR description please?

tcp/src/server.rs
```diff
@@ -88,7 +88,13 @@ impl<M: Metadata, S: Middleware<M> + 'static> ServerBuilder<M, S> {
 				let connections = SuspendableStream::new(listener.incoming());
 
 				let server = connections.map(move |socket| {
-					let peer_addr = socket.peer_addr().expect("Unable to determine socket peer address");
+					let peer_addr = match socket.peer_addr() {
+						Ok(addr) => addr,
+						Err(e) => {
+							trace!(target: "tcp", "Unable to determine socket peer address, ignoring connection {}", e);
```
dvdplm:
```suggestion
							warn!(target: "tcp", "Unable to determine socket peer address, ignoring connection {}", e);
```

PR #503, opened 2019-10-15
Integrate gitlab ci

fevo1971:


.gitlab-ci.yml
```diff
@@ -0,0 +1,52 @@
...
+  tags:
+    - rust-mac
+
+# check style 
+checkstyle-linux-stable:
+  image:                           parity/rust-builder:latest
+  stage:                           checkstyle
+  script:
+    - rustup component add rustfmt
+    - cargo fmt --all -- --check
```
dvdplm: Is jsonrpc using rustfmt? /cc @tomusdrw

tomusdrw: Yes, we used to enforce that on CI as well, but it was not working reliably on travis.

tomusdrw:
Thanks a lot @fevo1971 !!! Disabling travis / appveyor now.

PR #504, opened 2019-10-16
Fix race condition on `wait()`

foriequal0:
Manually dropping the future passed from `Future::select` before sending
`done_tx` prevents the race condition.

`Future::select` pass the unresolved future, which is a `Server` holding
`rpc_handler`, to the following callback. Therefore, it is dropped after
the `done_tx.send(())` after the callback exits.
It is possible that the thread that has executed `wait()`, which is
usually the main thread, terminates before the thread sending `done_tx`
drops `rpc_handler`.
Static variables are destructed at the end of termination of the main
thread, and then a segfault occurs when the thread dropping the
rpc_handler accesses the variables.

I closed the previous PR (https://github.com/paritytech/jsonrpc/pull/501).
Sorry for the confusion.

### DIFF
--- a/http/src/lib.rs
+++ b/http/src/lib.rs
@@ -566,7 +566,12 @@ fn serve<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>>(
 					}))
 					.map_err(|_| ())
 			})
-			.and_then(|_| done_tx.send(()))
+			.and_then(|(_, server)| {
+				// We drop the server first to prevent a situation where main thread terminates
+				// before the server is properly dropped (see #504 for more details)
+				drop(server);
+				done_tx.send(())
+			})
 	});
 }
 
--- a/ipc/src/server.rs
+++ b/ipc/src/server.rs
@@ -240,7 +240,10 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
 					.buffer_unordered(1024)
 					.for_each(|_| Ok(()))
 					.select(stop)
-					.map(|_| {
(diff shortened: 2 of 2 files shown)


http/src/lib.rs
```diff
@@ -565,7 +565,10 @@ fn serve<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>>(
 				}))
 				.map_err(|_| ())
 		})
-		.and_then(|_| done_tx.send(()))
+		.and_then(|(_, other)| {
+			drop(other);
```
tomusdrw: Could you add a comment what are we dropping here and why? Something like:

"We drop the `Server` first to prevent a situation where main thread terminates before the `Server` is properly dropped (see #504 for more details)".

ipc/src/server.rs
```diff
@@ -240,7 +240,8 @@ impl<M: Metadata, S: Middleware<M>> ServerBuilder<M, S> {
 					.buffer_unordered(1024)
 					.for_each(|_| Ok(()))
 					.select(stop)
-					.map(|_| {
+					.map(|(_, other)| {
```
tomusdrw: Same here if you don't mind.

foriequal0:
Thank you! I added comments to them.

http/src/lib.rs
dvdplm: maybe s/other/server/ then?

tomusdrw:
@foriequal0 please rebase as well while you address @dvdplm's grumble.

PR #507, opened 2019-10-16
Update syn & co

### LINKED ISSUE #505: Update syn of derive to v1.0
It's time to upgrade `syn`, `quote` and `proc-macro2` deps to v1.0 so that downstream users can gradually get rid of syn 0.15 in their crate graph.

tomusdrw:
Closes #505


derive/src/rpc_attr.rs
```diff
@@ -68,9 +68,15 @@ impl RpcMethodAttribute {
 	fn parse_meta(attr: &syn::Attribute, output: &syn::ReturnType) -> Option<Result<RpcMethodAttribute>> {
 		match attr.parse_meta().and_then(validate_attribute_meta) {
 			Ok(ref meta) => {
-				let attr_kind = match meta.name().to_string().as_ref() {
-					RPC_ATTR_NAME => Some(Self::parse_rpc(meta, output)),
-					PUB_SUB_ATTR_NAME => Some(Self::parse_pubsub(meta)),
+				let attr_kind = match meta
+					.path()
+					.get_ident()
+					.map(|i| i.to_string())
```
niklasad1: redundant closure 

```suggestion
					.map(ToString:to_string)
```

derive/src/to_delegate.rs
```diff
@@ -383,7 +380,7 @@ fn is_option_type(ty: &syn::Type) -> bool {
 		path.path
 			.segments
 			.first()
-			.map(|t| t.value().ident == "Option")
+			.map(|t| t.ident == "Option")
```
niklasad1: could be map_or

derive/src/rpc_attr.rs
```diff
@@ -275,3 +285,7 @@ fn get_aliases(ml: &syn::MetaList) -> Vec<String> {
 				.collect()
 		})
 }
+
+fn path_eq_str(path: &syn::Path, s: &str) -> bool {
+	path.get_ident().map(|i| i == s).unwrap_or(false)
```
niklasad1: could also be map_or

derive/src/rpc_attr.rs
```diff
@@ -162,24 +168,28 @@ fn validate_attribute_meta(meta: syn::Meta) -> Result<syn::Meta> {
...
 			}
 		}
 	}
 
 	let mut visitor = Visitor::default();
 	visit::visit_meta(&mut visitor, &meta);
 
-	match meta.name().to_string().as_ref() {
-		RPC_ATTR_NAME => {
+	let ident = meta.path().get_ident().map(|i| i.to_string());
```
niklasad1: would almost make sense the extract this pattern to separate function, `fn meta_into_str<'a>() -> Option<&'a str>` because it used a couple of times

tomusdrw:
@niklasad1 addressed :)

niklasad1:
@tomusdrw time to publish a new version? I would really want  to have #502, #504  and this `crates.io`

PR #512, opened 2019-10-23
Multiple methods to initiate same subscriptions

NikVolf:
@tomusdrw probably some integration test missing for this, but can't find where to place it


derive/tests/pubsub-macros.rs
```diff
@@ -49,6 +53,16 @@ impl Rpc for RpcImpl {
 		let _sink = subscriber.assign_id(SubscriptionId::Number(5));
 	}
 
+	fn subscribe_second(
+		&self,
+		_meta: Self::Metadata,
+		subscriber: Subscriber<String>,
```
tomusdrw: I wonder if we should enforce that the type of the subscription is the same as for the first one. Do you think it's possible to add?

NikVolf: yeah, addressed

tomusdrw: I actually meant to make sure that `Subscriber<String>` is used in both methods and we should return an error when we try to derive an implemenation that has the same subscription name and different types (for instance `Subscriber<String>` and `Subscriber<usize>`).

NikVolf: But isn't it up to implementor?

Or it will be no sound way to implement without such restriction?

tomusdrw: I think it's just going to be super confusing if we allow different types to be returned. I'd say that most likely it's going to be an error. The client of such subscription would need to expect one of the two kinds of the output actually (under the same subscription name).

NikVolf: Addressed as well!

Works locally, not sure how to show it with tests (it's just compilation error)

NikVolf: But it obviously can't tell that `std::String` and `String` are the same type, of course

tomusdrw: Yeah, that's fine. We have a set of high-level compilation-error tests here:
https://github.com/paritytech/jsonrpc/tree/master/derive/tests/ui
Feel free to add one here as well.

PR #518, opened 2019-11-15
Avoid using buffer_unordered

tomusdrw:
...and rather spawn tasks directly on the executor.

This should fix issues with unresponsive server, while still not breaking the previous fix causing the server to not be closed properly when there were pending requests. Now we use `Weak` references and only try to upgrade them when we handle the request.

### DIFF
--- a/http/src/handler.rs
+++ b/http/src/handler.rs
@@ -1,4 +1,4 @@
-use crate::Rpc;
+use crate::WeakRpc;
 
 use std::sync::Arc;
 use std::{fmt, mem, str};
@@ -16,7 +16,7 @@ use crate::{utils, AllowedHosts, CorsDomains, RequestMiddleware, RequestMiddlewa
 
 /// jsonrpc http request handler.
 pub struct ServerHandler<M: Metadata = (), S: Middleware<M> = middleware::Noop> {
-	jsonrpc_handler: Rpc<M, S>,
+	jsonrpc_handler: WeakRpc<M, S>,
 	allowed_hosts: AllowedHosts,
 	cors_domains: CorsDomains,
 	cors_max_age: Option<u32>,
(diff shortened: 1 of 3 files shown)


http/src/handler.rs
```diff
@@ -352,7 +352,11 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
 		}
 
 		// Read metadata
-		let metadata = self.jsonrpc_handler.extractor.read_metadata(&request);
+		let handler = match self.jsonrpc_handler.upgrade() {
```
dvdplm: Is it true that the only time we expect the upgrade to fail is when the tokio executor is closing?

http/src/lib.rs
```diff
@@ -553,11 +593,12 @@ fn serve<M: jsonrpc::Metadata, S: jsonrpc::Middleware<M>>(
 							keep_alive,
 						);
 
-						http.serve_connection(socket, service)
-							.map_err(|e| error!("Error serving connection: {:?}", e))
-							.then(|_| Ok(()))
+						tokio::spawn(
+							http.serve_connection(socket, service)
+								.map_err(|e| error!("Error serving connection: {:?}", e))
```
dvdplm: rustfmt is angry with you about this line. Maybe put the `map_err(|e| {` on the preceding line?

http/src/handler.rs
tomusdrw: Yes, since that's the only place where we keep a strong reference.

http/src/handler.rs
```diff
@@ -352,7 +352,11 @@ impl<M: Metadata, S: Middleware<M>> RpcHandler<M, S> {
...
+			Some(handler) => handler,
+			None => return RpcHandlerState::Writing(Response::closing()),
```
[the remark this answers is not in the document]
tomusdrw: Yeah, makes sense. Will update.

tomusdrw:
That one is returning an `application/json` content type, I need `plain`.

tomusdrw:
Merging, as @dvdplm reported no performance degradation.

PR #521, opened 2019-11-21
workaround serde deserializing incorrectly when arbitrary_precision enabled

### LINKED ISSUE #520: [jsonrpc-client-transports] jsonrpc does not deserialize if user has "arbitrary_precision" enabled in serde
If user is using `serde` with `features = ["arbitrary_precision"]` then jsonrpc errors:

```
ParseError("data did not match any variant of untagged enum ClientResponse", Error("data did not match any variant of untagged enum ClientResponse", line: 0, column: 0))
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Rpc(Other(Canceled))', src/libcore/result.rs:1187:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Related to: https://github.com/serde-rs/json/issues/505

jsonrpc deserializes into ClientResponse -> Output which are both `#[serde(untagged)]` enums

insipx:
serde deserializes incorrectly when arbitrary precision is enabled.

fixes #520 

The workaround is to deserialize into `Value` before `ClientResponse`. This is gated under the feature `arbitrary_precision` in core-client and `transports`.

### DIFF
--- a/core-client/transports/src/transports/mod.rs
+++ b/core-client/transports/src/transports/mod.rs
@@ -81,16 +81,27 @@ impl RequestBuilder {
 pub fn parse_response(
 	response: &str,
 ) -> Result<(Id, Result<Value, RpcError>, Option<String>, Option<SubscriptionId>), RpcError> {
-	serde_json::from_str::<ClientResponse>(&response)
-		.map_err(|e| RpcError::ParseError(e.to_string(), e.into()))
-		.map(|response| {
-			let id = response.id().unwrap_or(Id::Null);
-			let sid = response.subscription_id();
-			let method = response.method();
-			let value: Result<Value, Error> = response.into();
-			let result = value.map_err(RpcError::JsonRpcError);
-			(id, result, method, sid)
-		})
+	// https://github.com/serde-rs/json/issues/505
+	// Arbitrary precision confuses serde when deserializing into untagged enums,
+	// this is a workaround
+	let response = if cfg!(feature = "arbitrary_precision") {
+		let value = serde_json::from_str::<Value>(&response)
+			.map_err(|e| RpcError::ParseError(e.to_string(), e.into()))?;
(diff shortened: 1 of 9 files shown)


insipx:
there is now a generic fn `serde_from_str` which checks for `cfg!(feature = "arbitrary_precision")`.

Also made sure that cargo test --feature arbitrary_precision works

core-client/transports/src/transports/mod.rs
```diff
@@ -81,7 +81,7 @@ impl RequestBuilder {
 pub fn parse_response(
 	response: &str,
 ) -> Result<(Id, Result<Value, RpcError>, Option<String>, Option<SubscriptionId>), RpcError> {
-	serde_json::from_str::<ClientResponse>(&response)
+	jsonrpc_core::serde_from_str::<ClientResponse>(response)
```
tomusdrw: Can we now get rid of `serde_json` dependency?

core/src/lib.rs
```diff
@@ -54,3 +54,20 @@ pub use crate::io::{
 };
 pub use crate::middleware::{Middleware, Noop as NoopMiddleware};
 pub use crate::types::*;
+
+use serde_json::Error as SerdeError;
+
+/// workaround for https://github.com/serde-rs/json/issues/505
+/// Arbitrary precision confuses serde when deserializing into untagged enums,
+/// this is a workaround
+pub fn serde_from_str<'a, T>(input: &'a str) -> ::std::result::Result<T, SerdeError>
```
tomusdrw:
```suggestion
pub fn serde_from_str<'a, T>(input: &'a str) -> std::result::Result<T, SerdeError>
```

core-client/transports/src/transports/mod.rs
insipx: `serde_json` is still used for all `to_string` serializations

PR #527, opened 2019-12-04
New publish script.

tomusdrw:
Tag in one go, instead of crate-by-crate.
Don't fail in case tagging fails.

### DIFF
--- a/_automate/publish.sh
+++ b/_automate/publish.sh
@@ -25,7 +25,15 @@ for crate in ${ORDER[@]}; do
 	echo "Publishing $crate@$VERSION"
 	sleep 5
 	cargo publish $@ || read -p ">>>>> Publishing $crate failed. Press [enter] to continue. "
-	git tag -a "$crate-$VERSION" -m "$crate $VERSION"
+	cd -
+done
+
+# Make tags in one go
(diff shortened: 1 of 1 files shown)


_automate/publish.sh
```diff
@@ -25,7 +25,15 @@ for crate in ${ORDER[@]}; do
...
-	git tag -a "$crate-$VERSION" -m "$crate $VERSION"
+	cd -
+done
+
+# Make tags in one go
+for crate in ${ORDER[@]}; do
+	cd $crate
+	VERSION=$(grep "^version" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
+	echo "Tagging $crate@$VERSION"
+	git tag -a "$crate-$VERSION" -m "$crate $VERSION" || true
```
dvdplm: Don't you need to push the tags too?
```suggestion
	git tag -a "$crate-$VERSION" -m "$crate $VERSION" || true
	git push --tags
```

tomusdrw: We do `push --tags` after the loop. I don't want to push after every tag separately.

dvdplm: Ah I see, of course. :)

PR #530, opened 2020-01-06
[client] notify subscriber if subscription was rejected

### LINKED ISSUE #508: Subscription rejections are not handled correctly in `duplex.rs`
We just print a warning if we don't get expected `sid`, but instead we should propagate the error to the user so that they can see it.

https://github.com/paritytech/jsonrpc/blob/d1993a80176afdcbae6949bbf73ebbaf5625e644/core-client/transports/src/transports/duplex.rs#L217-L222

ascjones:
Closes #508 

- [x] test it works

### DIFF
--- a/core-client/transports/src/transports/duplex.rs
+++ b/core-client/transports/src/transports/duplex.rs
@@ -230,7 +230,7 @@ where
 									Ok(Async::NotReady) => {
 										self.incoming.push_back((id, result, Some(method), sid));
 										break;
-									},
+									}
 									Err(_) => {
 										log::warn!("aj-cli{}, but the reply channel has closed.", err);
 									}
--- a/core-client/transports/src/transports/duplex.rs
+++ b/core-client/transports/src/transports/duplex.rs
@@ -214,12 +214,27 @@ where
 									);
 								}
 							} else {
-								log::warn!(
-									"The response does not have expected subscription id in response: {:?} ({:?}): {:?}",
+								let err = RpcError::Other(format_err!(
+									"Subscription {:?} ({:?}) rejected: {:?}",
 									id,
 									method,
 									result,
-								);
+								));
+								match subscription.channel.poll_ready() {
+									Ok(Async::Ready(())) => {
+										subscription
+											.channel
+											.try_send(result)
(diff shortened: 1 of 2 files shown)


core-client/transports/src/transports/duplex.rs
```diff
@@ -214,12 +214,27 @@ where
...
+											.channel
+											.try_send(result)
+											.expect("The channel is ready; qed");
+									}
+									Ok(Async::NotReady) => {
+										self.incoming.push_back((id, result, Some(method), sid));
+										break;
+									}
+									Err(_) => {
+										log::warn!("aj-cli{}, but the reply channel has closed.", err);
```
tomusdrw: `aj-cli` looks funky :)

core-client/transports/src/transports/duplex.rs
```diff
@@ -214,12 +214,27 @@ where
...
+								));
+								match subscription.channel.poll_ready() {
+									Ok(Async::Ready(())) => {
+										subscription
+											.channel
+											.try_send(result)
+											.expect("The channel is ready; qed");
+									}
+									Ok(Async::NotReady) => {
+										self.incoming.push_back((id, result, Some(method), sid));
```
tomusdrw: What if it's the only request that is left? What are the chances that we will just loop forever on that item?

core-client/transports/src/transports/duplex.rs
ascjones: :sweat_smile:

core-client/transports/src/transports/duplex.rs
ascjones: It would loop as long as `subscription.channel` is `NotReady`, so as soon as the response channel is not full it would be added.

tomusdrw: Wouldn't that end up in busy-waiting for that channel to be ready? Also if we have single-threaded runtime it probably won't ever happen, since we are busy handling this `poll`, so we can't really process any of the responses.

ascjones: Wait I forgot about the `break` down there V V V

PR #534, opened 2020-01-22
Use tokio-compat for 0.2 timers in stdio crate

tomhoule:
Use-case: we rely on tokio 0.2 timers in our application code, but the stdio server crate runs the futures on the 0.1 executor. With tokio-compat, we should be able to get the best of both words, until `jsonrpc-core` works with `std::future::Future`.

### DIFF
--- a/stdio/Cargo.toml
+++ b/stdio/Cargo.toml
@@ -13,10 +13,10 @@ version = "14.0.1"
 futures = "0.1.23"
 jsonrpc-core = { version = "14.0", path = "../core" }
 log = "0.4"
-tokio = "0.1.7"
 tokio-codec = "0.1.0"
 tokio-io = "0.1.7"
 tokio-stdin-stdout = "0.1.4"
+tokio-compat = { version = "0.1.4", features = ["rt-full"] }
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


tomhoule:
I can't access the results of the failing CI check, so I don't know what's failing at the moment :)

stdio/Cargo.toml
```diff
@@ -11,12 +11,12 @@ version = "14.0.1"
 
 [dependencies]
 futures = "0.1.23"
-jsonrpc-core = { version = "14.0", path = "../core" }
```
tomusdrw: Please revert this change, we want the local crates to use local dependencies, not go to crates.io every time.

tomhoule:
Both changes made :)

tomhoule:
Update on this: we came up with an alternate solution with our own stdio adapter, feel free to merge or close this. For us to return to the jsonrpc-stdio-server crate it would need return a future instead of blocking, and be on tokio 0.2.

tomusdrw:
Transition to `tokio=0.2` and latest futures is a bigger task that will most likely change more things than this. I think it all should be addressed together.

PR #536, opened 2020-01-31
Fix ServerBuilder::cors_max_age doc comment time unit

jstarry:
#### Problem
The `cors_max_age` doc comment states that the age is specified in millis. But according to the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age), the header is specified in seconds.

#### Fix
Update doc comment to say "seconds"


jstarry:
[clabot:check]

jstarry:
Checkstyle fixed here: https://github.com/paritytech/jsonrpc/pull/537

jstarry:
@tomusdrw done!

tomusdrw:
Thanks!

PR #538, opened 2020-02-04
Fix a broken import in stdio example

sean0x42:
This is a minor fix which allows the example code in `stdio/README.md` to compile.


sean0x42:
[clabot:check]

tomusdrw:
Thanks!

PR #541, opened 2020-02-18
Client call with named params

### LINKED ISSUE #540: Support for by-name parameters in #[rpc(client)] macro
The RPC server I wish to communicate with only supports by-name parameters. I was hoping to build a nice typed adapter using the #[rpc(client)] macro but it appears this only supports by-position params. And in fact the `call_method` for `TypedClient` seems to ignore the fact that by-name calling is possible at all https://github.com/paritytech/jsonrpc/blob/ef44a62c543fd969dd5bedf5a4ece3f3814bb111/core-client/transports/src/lib.rs#L303

It seems like it would be pretty easy to add support for this with a switch in the proc macro to select the calling type. Is there a good reason I am missing as to why it hasn't been done this way? If not I'll submit a PR

willemolding:
# Add named_params option to RPC derive macro

Closes #540 

While the server generated by the derive macro doesn't support calling functions with named params some other systems do (some exclusively). It would still be nice to use the clean derive macro syntax to generate a client in this case.

This PR adds a flag `named_params` to the rpc method attribute. Example:
```rust
#[rpc(name = "add", named_params)]
fn add(&self, a: u32, b: u32) -> Result<String>;
```
In this case the json-rpc call object produced by calling the generated method on a client will have a params object that takes the key names from the function parameter names.
```json
{
   "jsonrpc": "2.0",
   "params": {"a": 4, "b": 2},
   ...
}
```

As the server doesn't support this it is important that this cannot be used when deriving server RPC methods. The rpc derive macro will throw a compile time error if the `named_params` switch is used in any case other than `#[rpc(client)]`. See (derive/tests/ui/attr-named-params-on-server.rs)

### DIFF
--- a/derive/src/options.rs
+++ b/derive/src/options.rs
@@ -64,7 +64,7 @@ impl DeriveOptions {
 		}
 		if options.enable_server && options.params_style == ParamStyle::Named {
 			// This is not allowed at this time
-			panic!("Server code generation only supports `params = \"positional\"` at this time. This is the default setting.")
+			panic!("Server code generation only supports `params = \"positional\"` (default) or `params = \"raw\" at this time.")
 		}
 		Ok(options)
 	}
--- a/derive/src/options.rs
+++ b/derive/src/options.rs
@@ -55,7 +55,8 @@ impl DeriveOptions {
 			options.enable_server = true;
 		}
 		if options.enable_server && options.params_style == ParamStyle::Named {
-			panic!()
+			// This is not allowed at this time
+			panic!("Server code generation only supports `params = \"positional\"` at this time. This is the default setting.")
 		}
 		Ok(options)
 	}
--- a/derive/src/options.rs
+++ b/derive/src/options.rs
@@ -54,6 +54,9 @@ impl DeriveOptions {
 			options.enable_client = true;
 			options.enable_server = true;
 		}
+		if options.enable_server && options.params_style == ParamStyle::Named {
+			panic!()
+		}
 		Ok(options)
 	}
 }
--- a/derive/src/options.rs
+++ b/derive/src/options.rs
@@ -28,8 +28,16 @@ impl DeriveOptions {
 		for arg in args {
 			if let syn::NestedMeta::Meta(meta) = arg {
 				match meta {
-					syn::Meta::Path(p) => {
-						match p.get_ident().unwrap().to_string().as_ref() {
+					syn::Meta::Path(ref p) => {
+						match p
+							.get_ident()
+							.ok_or(syn::Error::new_spanned(
+								p,
+								format!("Expecting identifier `{}` or `{}`", CLIENT_META_WORD, SERVER_META_WORD),
+							))?
+							.to_string()
+							.as_ref()
+						{
 							CLIENT_META_WORD => options.enable_client = true,
 							SERVER_META_WORD => options.enable_server = true,
 							_ => {}
--- a/derive/src/options.rs
+++ b/derive/src/options.rs
@@ -1,37 +1,59 @@
-use proc_macro::TokenStream;
+use std::str::FromStr;
+
+use crate::params_style::ParamStyle;
+use crate::rpc_attr::path_eq_str;
+
+const CLIENT_META_WORD: &str = "client";
+const SERVER_META_WORD: &str = "server";
+const PARAMS_META_KEY: &str = "params";
 
 #[derive(Debug)]
 pub struct DeriveOptions {
 	pub enable_client: bool,
 	pub enable_server: bool,
+	pub params_style: ParamStyle,
 }
 
 impl DeriveOptions {
-	pub fn new(enable_client: bool, enable_server: bool) -> Self {
+	pub fn new(enable_client: bool, enable_server: bool, params_style: ParamStyle) -> Self {
 		DeriveOptions {
 			enable_client,
 			enable_server,
+			params_style,
 		}
 	}
 
-	pub fn try_from(tokens: TokenStream) -> Result<Self, syn::Error> {
-		if tokens.is_empty() {
-			return Ok(Self::new(true, true));
-		}
-		let ident: syn::Ident = syn::parse::<syn::Ident>(tokens)?;
-		let options = {
-			let ident = ident.to_string();
-			if ident == "client" {
-				Some(Self::new(true, false))
-			} else if ident == "server" {
-				Some(Self::new(false, true))
-			} else {
-				None
+	pub fn try_from(args: syn::AttributeArgs) -> Result<Self, syn::Error> {
+		let mut options = DeriveOptions::new(false, false, ParamStyle::default());
+		for arg in args {
+			if let syn::NestedMeta::Meta(meta) = arg {
+				match meta {
+					syn::Meta::Path(p) => {
+						match p.get_ident().unwrap().to_string().as_ref() {
+							CLIENT_META_WORD => options.enable_client = true,
+							SERVER_META_WORD => options.enable_server = true,
+							_ => {}
+						};
+					}
+					syn::Meta::NameValue(nv) => {
+						if path_eq_str(&nv.path, PARAMS_META_KEY) {
+							if let syn::Lit::Str(ref lit) = nv.lit {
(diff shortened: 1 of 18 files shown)


willemolding:
[clabot:check]

derive/tests/client.rs
```diff
@@ -3,43 +3,91 @@ use jsonrpc_core::{IoHandler, Result};
...
-			eprintln!("{:?}", err);
-			assert!(false);
+mod named_params {
+	use super::*;
+	use jsonrpc_core::Params;
+	use serde_json::json;
+
+	#[rpc(client)]
+	pub trait Rpc {
+		#[rpc(name = "call_with_named", named_params)]
```
ascjones: Would it make sense to specify `named_params` in the trait level attribute instead? On the one hand it's good to have the fine grained control - but in practice wouldn't all methods be either named or positional?

tomusdrw: We already have `raw_params` on the method level. I think this is a good idea to have it here, but an additional attribute on the trait level would be cool tool, especially in the future if we implement server side as well.

willemolding:
I like the idea of having it at the trait level to set the default, which can then be overridden at the method level. That way we have the fined grained control as well. I'm picturing something like 

```rust
#[rpc(client, params = "named")] // sets the default to named params
pub trait Rpc {
	#[rpc(name = "call_with_named")] // uses the trait default
        fn call_with_named() -> {
        ...
        #[rpc(name = "call_positional", params = "positional")] // overrides trait default
        fn call_positional() -> {
}

```

tomusdrw:
That looks really nice, are you willing to change how `raw_params` work as well, I imagine that `params` would take any of `"raw"|"positional"|"named"`.

I guess we would need to keep `raw_params` for one version and mark it deprecated (i.e produce a warning if used).

willemolding:
Ok getting pretty close.

There are now three options for params: `["named", "positional", "raw"]`. You can also use the old `raw_params` flag which overrides the setting but I couldn't figure out how to trigger a compiler warning.

This can be set at the trait level (which then becomes the default for all trait methods without a `params` attribute key. It can also be used at the method level which overrides the default from the trait so both calling type can be used together.

I'm done for today but I would appreciate suggestions for any more tests this might need to be sure it is solid

derive/src/rpc_attr.rs
```diff
@@ -9,7 +11,7 @@ pub struct RpcMethodAttribute {
 	pub name: String,
 	pub aliases: Vec<String>,
 	pub kind: AttributeKind,
-	pub raw_params: bool,
+	pub params_style: Option<ParamStyle>, // None means do not overrite the top level default
```
willemolding:
```suggestion
	pub params_style: Option<ParamStyle>, // None means do not override the top level default
```

derive/src/options.rs
```diff
@@ -1,37 +1,63 @@
...
-				Some(Self::new(false, true))
-			} else {
-				None
+	pub fn try_from(args: syn::AttributeArgs) -> Result<Self, syn::Error> {
+		let mut options = DeriveOptions::new(false, false, ParamStyle::default());
+		for arg in args {
+			if let syn::NestedMeta::Meta(meta) = arg {
+				match meta {
+					syn::Meta::Path(p) => {
+						match p.get_ident().unwrap().to_string().as_ref() {
```
tomusdrw: Can this `unwrap` fail? Can we return an error instead?

derive/src/options.rs
```diff
@@ -1,37 +1,63 @@
...
-			None => Err(syn::Error::new(ident.span(), "Unknown attribute.")),
 		}
+		if !options.enable_client && !options.enable_server {
+			// if nothing provided default to both
+			options.enable_client = true;
+			options.enable_server = true;
+		}
+		if options.enable_server && options.params_style == ParamStyle::Named {
+			// This is not allowed at this time
+			panic!("Server code generation only supports `params = \"positional\"` at this time. This is the default setting.")
```
tomusdrw: I guess it also supports `params = "raw"`, no?

derive/src/rpc_attr.rs
```diff
@@ -81,12 +84,18 @@ impl RpcMethodAttribute {
 								let aliases = get_meta_list(&meta).map_or(Vec::new(), |ml| get_aliases(ml));
 								let raw_params =
 									get_meta_list(meta).map_or(false, |ml| has_meta_word(RAW_PARAMS_META_WORD, ml));
+								let params_style = match raw_params {
+									true => Ok(Some(ParamStyle::Raw)),
```
tomusdrw: Can we print a deprecation warning here?

derive/src/rpc_trait.rs
```diff
@@ -218,13 +223,19 @@ fn rpc_wrapper_mod_name(rpc_trait: &syn::ItemTrait) -> syn::Ident {
 	syn::Ident::new(&mod_name, proc_macro2::Span::call_site())
 }
 
+fn has_named_params(methods: &[RpcMethod]) -> bool {
+	methods
+		.iter()
+		.any(|method| method.attr.params_style == std::option::Option::Some(ParamStyle::Named))
```
tomusdrw:
```suggestion
		.any(|method| method.attr.params_style == Some(ParamStyle::Named))
```

derive/src/to_delegate.rs
```diff
@@ -238,7 +238,7 @@ impl RpcMethod {
 				self.params_with_trailing(trailing_args_num, param_types, tuple_fields)
 			} else if param_types.is_empty() {
 				quote! { let params = params.expect_no_params(); }
-			} else if self.attr.raw_params {
+			} else if self.attr.params_style == Some(ParamStyle::Raw) {
 				quote! { let params: _jsonrpc_core::Result<_> = Ok((params,)); }
 			} else {
 				quote! { let params = params.parse::<(#(#param_types, )*)>(); }
```
tomusdrw: I guess this is where we should add the named style server-side implementation, right?

derive/src/rpc_attr.rs
willemolding: Unfortunately I can't figure out a way to emit a warning without using the `proc_macro::Diagnostic` feature which is unstable

Not sure if we have any other ideas for this or if it is ok to just add a depreciation note to the docs but not the macro

derive/src/options.rs
willemolding: Throws a nice compiler error now instead

derive/src/options.rs
willemolding: Nice catch

derive/src/to_delegate.rs
willemolding: Added an `unimplemented!` as a helper to someone implementing server side named params in the future

willemolding:
Thanks for the review @tomusdrw. The only thing lacking is a deprecation warning for using the now-deprecated `raw_params` flag. Not sure if anyone knows a trick for this or are we happy to just have the updated docs showing the new way?

derive/src/rpc_attr.rs
tomusdrw: I think in the past we simply used to `println` the warning. @ascjones do you recall?

willemolding:
It looks like using `println!` won't make a compile time error/warning. It would be a matter of making the generated code contain a runtime `println!` which seems wrong to me.

I left a placeholder warning so if/when the proc-macro diagnostics becomes stable it will be simple to add it in.

tomusdrw:
Apparently it seems to be upstream issue, could you fix the `quote` version like here? https://github.com/paritytech/substrate/pull/5135

willemolding:
> Apparently it seems to be upstream issue, could you fix the `quote` version like here? [paritytech/substrate#5135](https://github.com/paritytech/substrate/pull/5135)

Hmm this doesn't seem to have fixed it. It sure is a pain not being able to see the CI output

tomusdrw:
> Hmm this doesn't seem to have fixed it. It sure is a pain not being able to see the CI output

Yeah, sorry about that. Seems that `1.0.2` is yanked. Could you roll back to `=1.0.1`? I tested that locally and it should work.

tomusdrw:
Thanks!

Issue #542, opened 2020-02-19
Add "Accept" Header for http transport for client


yinhu-prismalab:
I am not sure about whether it should be there or not by protocol/standard but it seems some existing libs for other language (e.g. http-jsonrpc-server for nodejs) requires "application/json" for "Accept" header alone with the "Content-type" the current implementation have.

It would be nice to have a way to  config the headers not only for working with other libs but also exiting old systems that one cannot modify.

tomusdrw:
I believe we do allow `Accept` header to be sent on the requests, do you mean it should be included in the response as well? This header is request-only, I don't think that would work.

Could you point me to some examples what kind of requests/responses the client library sends/expects?

yinhu-prismalab:
I mean the client part: jsonrpc/core-client/transports/src/transports/http.rs , it seems only content-type is set. My current code is based on an older version of this crate, I haven't tried latest version yet. But the code on github seems the still the same.

tomusdrw:
@yinhu-prismalab Oh, I see. Yeah, we should definitely send `Accept` header unconditionally in the client. Feel free to submit a PR, happy to review it.

yinhu-prismalab:
Created PR #544 . 
Simply  added a new header in the requests.
Please let me know if I should modify somewhere else as well.

tomusdrw:
Resolved via #544

PR #544, opened 2020-02-21
Add "Accept:application/json" header for all requests.

yinhu-prismalab:
As discussed in #542 .
Add accept header for client http requests so that it can work with some other libs/existing servers that require this field.


tomusdrw:
Looks good, thanks!

yinhu-prismalab:
[clabot:check]

PR #545, opened 2020-04-14
14.1.0 bump

ascjones:
## Release notes

- Wait for http server to close on Drop. (#543)
- Add "Accept:application/json" header for all requests. (#544) 
- Client call with named params (#541)


ascjones:
Published

PR #546, opened 2020-04-14
Wait for published version to be available

ascjones:
When I was publishing `14.1.0` the publish script repeatedly failed where a dependency on a previously published crate was not yet available.

This PR waits for the published version to be available after publishing each package before moving on to the next one.

Note that I have only tested this with the `cargo publish` commented out, not with an actual release.


ascjones:
> Hmm actually, let's leave sleep 5 too. I think it was there to be able to CTRL+C in case you figure out there is something wrong.

Ah sorry thought it was an arbitrary sleep to wait for the package to be available

dvdplm:
> Hmm actually, let's leave `sleep 5` too. I think it was there to be able to `CTRL+C` in case you figure out there is something wrong.

Feature request: set up a Ctrl+C hook, and auto-yank previous items! ;)

PR #548, opened 2020-05-19
Add a subscription manager

HCastano:
I was implementing some pubsub APIs in Substrate, outside of `sc-rpc`, and got to a point where I needed something to execute the subscription Futures. There's nothing provided out of the box for doing that, so after talking with @tomusdrw we figured that it might be useful to have the [Substrate subscription manager](https://github.com/paritytech/substrate/blob/master/client/rpc-api/src/subscriptions.rs) be a part of the `pubsub` crate.

What I've done is I've stolen that code and changed it to try and be a little less opinionated. One concern I have though is that I've changed `SubscriptionId` to now take a generic parameter, `N`, to avoid only working with `u64` ids, and I think this will break people's code.

I'm looking for feedback on whether the general approach is alright, and if the change to `SubscriptionId` is acceptable.

### DIFF
--- a/pubsub/src/manager.rs
+++ b/pubsub/src/manager.rs
@@ -130,12 +130,12 @@ impl Default for RandomStringIdProvider {
 pub struct SubscriptionManager<I: IdProvider = RandomStringIdProvider> {
 	id_provider: I,
 	active_subscriptions: ActiveSubscriptions,
-	executor: TaskExecutor, // Make generic?
+	executor: TaskExecutor,
 }
 
 impl<I: IdProvider> SubscriptionManager<I> {
 	/// Creates a new SubscriptionManager.
-	pub fn new_with_id_provider(id_provider: I, executor: TaskExecutor) -> Self {
+	pub fn with_id_provider(id_provider: I, executor: TaskExecutor) -> Self {
 		Self {
 			id_provider,
 			active_subscriptions: Default::default(),
--- a/pubsub/src/manager.rs
+++ b/pubsub/src/manager.rs
@@ -1,7 +1,6 @@
 //! Provides an executor for subscription Futures.
 
 use std::collections::HashMap;
-use std::hash::Hash;
 use std::sync::Arc;
 
 use crate::core::futures::sync::oneshot;
@@ -19,7 +18,7 @@ pub type TaskExecutor = Arc<dyn future::Executor<Box<dyn Future<Item = (), Error
 /// Trait used to provide unique subscription ids.
 pub trait IdProvider {
 	/// A unique ID used to identify a subscription.
-	type Id: Copy + Clone + Default + Eq + Hash + Into<SubscriptionId> + From<SubscriptionId>;
+	type Id: Copy + Clone + Default + Into<SubscriptionId>;
 
 	/// Returns next id for the subscription.
 	fn next_id(&self) -> Self::Id;
@@ -32,7 +31,7 @@ pub trait IdProvider {
 #[derive(Clone)]
 pub struct SubscriptionManager<I: Default + IdProvider> {
 	next_id: I,
-	active_subscriptions: Arc<Mutex<HashMap<I::Id, oneshot::Sender<()>>>>,
+	active_subscriptions: Arc<Mutex<HashMap<SubscriptionId, oneshot::Sender<()>>>>,
 	executor: TaskExecutor, // Make generic?
 }
 
@@ -65,7 +64,7 @@ impl<I: Default + IdProvider> SubscriptionManager<I> {
 				.select(rx.map_err(|e| warn!("Error timing out: {:?}", e)))
 				.then(|_| Ok(()));
 
-			self.active_subscriptions.lock().insert(id, tx);
+			self.active_subscriptions.lock().insert(subscription_id.clone(), tx);
 			if self.executor.execute(Box::new(future)).is_err() {
 				error!("Failed to spawn RPC subscription task");
 			}
@@ -78,7 +77,7 @@ impl<I: Default + IdProvider> SubscriptionManager<I> {
 	///
 	/// Returns true if subscription existed or false otherwise.
 	fn cancel(&self, id: SubscriptionId) -> bool {
-		if let Some(tx) = self.active_subscriptions.lock().remove(&id.into()) {
+		if let Some(tx) = self.active_subscriptions.lock().remove(&id) {
 			let _ = tx.send(());
 			return true;
 		}
--- a/pubsub/src/manager.rs
+++ b/pubsub/src/manager.rs
@@ -184,7 +184,6 @@ impl<I: IdProvider> SubscriptionManager<I> {
 				.then(|_| Ok(()));
 
 			self.active_subscriptions.lock().insert(subscription_id.clone(), tx);
-			dbg!(&self.active_subscriptions.lock().get(&subscription_id));
 			if self.executor.execute(Box::new(future)).is_err() {
 				error!("Failed to spawn RPC subscription task");
 			}
@@ -197,7 +196,6 @@ impl<I: IdProvider> SubscriptionManager<I> {
 	///
 	/// Returns true if subscription existed or false otherwise.
 	pub fn cancel(&self, id: SubscriptionId) -> bool {
-		dbg!(&self.active_subscriptions.lock().get(&id));
 		if let Some(tx) = self.active_subscriptions.lock().remove(&id) {
 			let _ = tx.send(());
 			return true;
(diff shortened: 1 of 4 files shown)


pubsub/src/manager.rs
```diff
@@ -0,0 +1,115 @@
...
+};
+use log::{error, warn};
+use parking_lot::Mutex;
+
+/// Alias for an implementation of `futures::future::Executor`.
+pub type TaskExecutor = Arc<dyn future::Executor<Box<dyn Future<Item = (), Error = ()> + Send>> + Send + Sync>;
+
+/// Trait used to provide unique subscription ids.
+pub trait IdProvider {
+	// TODO: Maybe have this impl Into<u64>?
```
tomusdrw, line 24: It should be possible to have `String` ids as well.

pubsub/src/manager.rs
```diff
@@ -0,0 +1,115 @@
...
+
+/// Subscriptions manager.
+///
+/// Takes care of assigning unique subscription ids and
+/// driving the sinks into completion.
+#[derive(Clone)]
+pub struct Manager<I: Default + IdProvider> {
+	next_id: I,
+	active_subscriptions: Arc<Mutex<HashMap<I::Id, oneshot::Sender<()>>>>,
+	executor: TaskExecutor, // Make generic?
```
tomusdrw, line 63: IMHO It's fine, in most workloads spawning subscription is not going to be a bottleneck so we can safely do a virtual call here.

pubsub/src/manager.rs
```diff
@@ -0,0 +1,115 @@
...
+pub trait IdProvider {
+	// TODO: Maybe have this impl Into<u64>?
+	type Id: Clone + Default + Eq + Hash;
+
+	/// Returns next id for the subscription.
+	fn next_id(&self) -> Self::Id;
+}
+
+/// Trait used to drive subscription Futures to completion.
+pub trait SubscriptionManager {
```
tomusdrw, line 32: I'm not sure the trait is really needed, I'd rather have `Manager` with a default generic parameter of `u64` subscription `IdProvider`. Mainly to avoid importing extra traits to call functions.
Do you have some use cases in mind where having it behind trait would be useful?

pubsub/src/types.rs
```diff
@@ -35,47 +36,68 @@ impl<T: PubSubMetadata> PubSubMetadata for Option<T> {
 }
 
 /// Unique subscription id.
+///
 /// NOTE Assigning same id to different requests will cause the previous request to be unsubscribed.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
-pub enum SubscriptionId {
-	/// U64 number
-	Number(u64),
+pub enum SubscriptionId<N> {
```
tomusdrw, line 42: Ah, I see what you meant with `Into<u64>` in the earlier comment. Why not require `Into<SubscriptionId>` instead? This together with some extra `From` implementations that we would add here (for numeric types and `String` as you did now) would allow us to keep that configurable without maiking `SubscriptionId` generic.

pubsub/src/manager.rs
HCastano: Yeah, my reasoning kinda was that if people wanted a "default" implementation of a `SubscriptionManager` they'd use the `Manager` struct, and if they wanted to roll their own they just had to conform to the `SubscriptionManager` trait. Is it reasonable to expect that people will want to use something other than the "default" implementation? If not, then yeah we don't need to the trait.

tomusdrw: Since the trait is not really used in the library at all I think a concrete implementation is enough. If someone is building their own solution and needs that abstraction they can introduce a trait themselves and implement it for the type from our library.

pubsub/src/manager.rs
```diff
@@ -0,0 +1,88 @@
+//! Provides an executor for subscription Futures.
```
tomusdrw, line 1: I think we should expand on this module description a bit. The subscription manager is opinionated way to handle processing subscriptions that are based on `Stream` coming from the user codebase.
The manager takes care of:
1. Assigning the ID
2. Consuming a `Stream` of events (coming from user code) and transforming it into a subscription notifications to the client.
3. Spawning a resulting future to executor
4. Cancelling the stream when the subscription is canceled.

pubsub/src/manager.rs
```diff
@@ -0,0 +1,88 @@
...
+/// Takes care of assigning unique subscription ids and
+/// driving the sinks into completion.
+#[derive(Clone)]
+pub struct SubscriptionManager<I: Default + IdProvider> {
+	next_id: I,
+	active_subscriptions: Arc<Mutex<HashMap<I::Id, oneshot::Sender<()>>>>,
+	executor: TaskExecutor, // Make generic?
+}
+
+impl<I: Default + IdProvider> SubscriptionManager<I> {
```
tomusdrw, line 39: `Default` should only be required for `new` (not for `add/executor/etc`) and there should be an extra method `with_id_provider(TaskExecutor, IdProvider)` for non-default cases.

pubsub/src/manager.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// Subscriptions manager.
+///
+/// Takes care of assigning unique subscription ids and
+/// driving the sinks into completion.
+#[derive(Clone)]
+pub struct SubscriptionManager<I: Default + IdProvider> {
+	next_id: I,
+	active_subscriptions: Arc<Mutex<HashMap<I::Id, oneshot::Sender<()>>>>,
```
tomusdrw, line 35:
```suggestion
	active_subscriptions: Arc<Mutex<HashMap<SubscriptionId, oneshot::Sender<()>>>>,
```

IMHO using `SubscriptionId` will greately simply the `Id` bounds and I think it doesn't really limit the functionality at all.

pubsub/src/manager.rs
```diff
@@ -0,0 +1,88 @@
...
+	/// Returns next id for the subscription.
+	fn next_id(&self) -> Self::Id;
+}
+
+/// Subscriptions manager.
+///
+/// Takes care of assigning unique subscription ids and
+/// driving the sinks into completion.
+#[derive(Clone)]
+pub struct SubscriptionManager<I: Default + IdProvider> {
```
tomusdrw, line 33:
```suggestion
pub struct SubscriptionManager<I: IdProvider = RandomizedStringProvider> {
```

We should implement two `IdProviders`:
1. `NumericIdProvider`
2. `RandomizedStringProvider`

The first one would simply use `AtomicU64` and would return increasing numeric values, the second one - which should also be a default to use if `SubscriptionManager` does not include generic parameter - should return a randomized strings (guids).
The reason for using randomized strings is to make sure that subscription ids are not easy to guess, so that multiple clients can't cancel each-others subscriptions easily. The first one will be mostly usable for tests or cases where the aforementioned issue is not a problem.

pubsub/src/types.rs
```diff
@@ -76,30 +71,83 @@ impl From<SubscriptionId> for core::Value {
...
+
+	#[test]
+	fn should_convert_from_u64_to_subscription_id() {
+		let val = 5u64;
+		let res: SubscriptionId = val.into();
+		assert_eq!(res, SubscriptionId::Number(5));
+	}
+
+	#[test]
+	fn should_convert_from_string_to_subscription_id() {
```
tomusdrw, line 148: Thanks for all the test :heart:

pubsub/src/manager.rs
```diff
@@ -1,62 +1,166 @@
...
+
+impl IdProvider for RandomStringIdProvider {
+	type Id = String;
+
+	fn next_id(&self) -> Self::Id {
+		let mut rng = thread_rng();
+		let id: String = iter::repeat(())
+			.map(|()| rng.sample(Alphanumeric))
+			.take(self.len)
+			.collect();
```
tomusdrw, line 114: Nice and elegant :)

pubsub/src/manager.rs
```diff
@@ -1,62 +1,166 @@
...
 /// Takes care of assigning unique subscription ids and
 /// driving the sinks into completion.
 #[derive(Clone)]
-pub struct SubscriptionManager<I: Default + IdProvider> {
-	next_id: I,
-	active_subscriptions: Arc<Mutex<HashMap<I::Id, oneshot::Sender<()>>>>,
+pub struct SubscriptionManager<I: IdProvider = RandomStringIdProvider> {
+	id_provider: I,
+	active_subscriptions: ActiveSubscriptions,
 	executor: TaskExecutor, // Make generic?
```
tomusdrw, line 133:
```suggestion
	executor: TaskExecutor, 
```

pubsub/src/manager.rs
```diff
@@ -1,62 +1,166 @@
...
 }
 
-impl<I: Default + IdProvider> SubscriptionManager<I> {
+impl<I: IdProvider> SubscriptionManager<I> {
 	/// Creates a new SubscriptionManager.
-	pub fn new(executor: TaskExecutor) -> Self {
+	pub fn new_with_id_provider(id_provider: I, executor: TaskExecutor) -> Self {
```
tomusdrw, line 138:
```suggestion
	pub fn with_id_provider(id_provider: I, executor: TaskExecutor) -> Self {
```
AFAIK the overall convention is to skip `new_` in `with_` constructors.

pubsub/src/types.rs
```diff
@@ -62,6 +62,14 @@ impl From<String> for SubscriptionId {
 	}
 }
 
+// TODO: Don't unwrap, and probably implement TryFrom instead of From
+use std::convert::TryInto;
+impl From<usize> for SubscriptionId {
```
tomusdrw, line 67: I think instead of providing this `From<usize>` implementation we should rather change the `NumericIdProvider` to convert `usize` to `u64` internally. The reason is that it's highly unlikely for `NumericIdProvider` to reach the case where this conversion might fail (given it runs on 128-bit processor anyway O.o) and here we are making stronger assumption that `usize` can be safely converted to `SubscriptionId`.

pubsub/src/manager.rs
HCastano: Unfortunately I can't take credit for it: https://rust-random.github.io/rand/rand/distributions/struct.Alphanumeric.html

HCastano:
I want to write some tests to make sure the `SubscriptionManager` can `add()` and `cancel()` subscriptions before I mark this as "ready for review" (even though you've already reviewed it @tomusdrw , lol).

HCastano:
Looks like the Windows runner is failing due to an old `rustc` version which doesn't have `matches!` stabilized yet

PR #557, opened 2020-06-03
Allow skipping and retrying publishing.

tomusdrw:
Towards idempotency :dancing_men: 

I think the next reasonable change would be to check version before publish and if it matches, skip by default, but propose to re-publish (maybe not even needed)?
But I had enough bash programming for today :)

The `set -x/+x` has been sprinkled to clean up the output a bit - I still like to see the commands that are actually made, but the looping/conditionals where causing havoc in the logs.

### DIFF
--- a/_automate/publish.sh
+++ b/_automate/publish.sh
@@ -27,7 +27,7 @@ for crate in ${ORDER[@]}; do
 		VERSION=$(grep "^version" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
 		# give the user an opportunity to abort or skip before publishing
 		RET=""
-		read -t 5 -p "Publishing $crate@$VERSION. Type [s] to skip. " RET || true
+		read -t 5 -p "Publishing $crate@$VERSION. Type [s] to skip, or any key to proceed. " RET || true
 		if [ "$RET" != "s" ]; then
 			set -x
 			cargo publish $@ || read -p ">>>>> Publishing $crate failed. Press [enter] to continue or type [r] to retry. " CHOICE
--- a/_automate/publish.sh
+++ b/_automate/publish.sh
@@ -86,6 +86,10 @@ for CRATE_DIR in ${ORDER[@]}; do
 done
 
 # Make tags in one go
+set -x
+git fetch --tags
+set +x
+
 for CRATE_DIR in ${ORDER[@]}; do
 	cd $CRATE_DIR > /dev/null
 	read_toml
--- a/_automate/publish.sh
+++ b/_automate/publish.sh
@@ -39,7 +39,7 @@ for CRATE_DIR in ${ORDER[@]}; do
 	# Seems the latest version matches, skip by default.
 	if [ "$REMOTE_VERSION" = "$VERSION" ] || [[ "$REMOTE_VERSION" > "$VERSION" ]]; then
 		RET=""
-		echo "Seems that $NAME@$REMOTE_VERSION is available. Continuing in 5s. "
+		echo "Seems like $NAME@$REMOTE_VERSION is already published. Continuing in 5s. "
 		read -t 5 -p ">>>> Type [r][enter] to retry, or [enter] to continue... " RET || true
 		if [ "$RET" != "r" ]; then
 			echo "Skipping $NAME@$VERSION"
@@ -69,13 +69,13 @@ for CRATE_DIR in ${ORDER[@]}; do
 		fi
 	done
 
-	# Wait again to make sure the published version is already available.
-	echo "Waiting for $NAME@$VERSION to be available..."
+	# Wait again to make sure that the new version is published and available.
+	echo "Waiting for $NAME@$VERSION to become available at the registry..."
 	while : ; do
 		sleep 3
 		remote_version
 		if [ "$REMOTE_VERSION" = "$VERSION" ]; then
-			echo "🥳 $NAME@$VERSION published succesfuly."
+			echo "🥳 $NAME@$VERSION published succesfully."
 			sleep 3
 			break
 		else
--- a/_automate/publish.sh
+++ b/_automate/publish.sh
@@ -18,13 +18,29 @@ set -x
 
 cargo clean
 
+set +x
+
 # Then actually perform publishing.
 for crate in ${ORDER[@]}; do
 	cd $crate
-	VERSION=$(grep "^version" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
-	echo "Publishing $crate@$VERSION"
-	sleep 5 # give the user an opportunity to abort before publishing
-	cargo publish $@ || read -p ">>>>> Publishing $crate failed. Press [enter] to continue. "
+	while : ; do
+		VERSION=$(grep "^version" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
+		# give the user an opportunity to abort or skip before publishing
+		RET=""
+		read -t 5 -p "Publishing $crate@$VERSION. Type [s] to skip. " RET || true
+		if [ "$RET" != "s" ]; then
+			set -x
+			cargo publish $@ || read -p ">>>>> Publishing $crate failed. Press [enter] to continue or type [r] to retry. " CHOICE
+			set +x
+			if [ "$CHOICE" != "r" ]; then
+				break
+			fi
+		else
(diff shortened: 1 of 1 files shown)


_automate/publish.sh
```diff
@@ -18,13 +18,29 @@ set -x
...
 	cd $crate
-	VERSION=$(grep "^version" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
-	echo "Publishing $crate@$VERSION"
-	sleep 5 # give the user an opportunity to abort before publishing
-	cargo publish $@ || read -p ">>>>> Publishing $crate failed. Press [enter] to continue. "
+	while : ; do
+		VERSION=$(grep "^version" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
+		# give the user an opportunity to abort or skip before publishing
+		RET=""
+		read -t 5 -p "Publishing $crate@$VERSION. Type [s] to skip. " RET || true
```
dvdplm, line 30:
```suggestion
		read -t 5 -p "Publishing $crate@$VERSION. Type [s] to skip, or any key to proceed. " RET || true
```
maybe?

_automate/publish.sh
```diff
@@ -18,13 +18,29 @@ set -x
...
+		if [ "$RET" != "s" ]; then
+			set -x
+			cargo publish $@ || read -p ">>>>> Publishing $crate failed. Press [enter] to continue or type [r] to retry. " CHOICE
```
dvdplm, line 33: "any key to continue" here too, or print `[enter] to continue` above as well?

tomusdrw:
> When RIIR? ;)

Bash > Rust, we should consider rewriting the whole library.

Change the script a bit more - it now skips by default if the remote version is already released. Please take a look again.

_automate/publish.sh
```diff
@@ -21,54 +32,80 @@ cargo clean
...
-for crate in ${ORDER[@]}; do
-	cd $crate
+for CRATE_DIR in ${ORDER[@]}; do
+	cd $CRATE_DIR > /dev/null
+	read_toml
+	remote_version
+	# Seems the latest version matches, skip by default.
+	if [ "$REMOTE_VERSION" = "$VERSION" ] || [[ "$REMOTE_VERSION" > "$VERSION" ]]; then
+		RET=""
+		echo "Seems that $NAME@$REMOTE_VERSION is available. Continuing in 5s. "
```
dvdplm, line 42:
```suggestion
		echo "Seems like $NAME@$REMOTE_VERSION is already published. Continuing in 5s. "
```

_automate/publish.sh
```diff
@@ -4,11 +4,22 @@ set -eu
...
+function read_toml () {
+	NAME=""
+	VERSION=""
+	NAME=$(grep "^name" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
 	VERSION=$(grep "^version" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
-	echo "$crate@$VERSION"
+}
+function remote_version () {
+	REMOTE_VERSION=""
+	REMOTE_VERSION=$(cargo search "$NAME" | grep "^$NAME =" | sed -e 's/.*"\(.*\)".*/\1/')
```
dvdplm, line 15: `'s/.*"\(.*\)".*/\1/'` doesn't handle `alpha`/`beta` crates but I don't think we care (I also don't know if those even show up in `cargo search`?).

Another nit would be pathological cases such as `cargo search rust` (18676 matches): only the first 10 are displayed (no idea what ordering criteria, maybe the same as crates.io?) so if the one you're publishing is further down the list you'd need to use `--limit <some int>`. Again, I don't think that is important but you asked for a review so… ;)

_automate/publish.sh
```diff
@@ -21,54 +32,80 @@ cargo clean
...
 
-  echo "  Waiting for published version $VERSION to be available..."
-	CRATE_NAME=$(grep "^name" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
-	LATEST_VERSION=0
-	while [[ $LATEST_VERSION != $VERSION ]]
-	do
-	  sleep 3
-	  LATEST_VERSION=$(cargo search "$CRATE_NAME" | grep "^$CRATE_NAME =" | sed -e 's/.*"\(.*\)".*/\1/')
-	  echo "    Latest available version: $LATEST_VERSION"
+	# Wait again to make sure the published version is already available.
```
dvdplm, line 72:
```suggestion
	# Wait again to make sure that the new version is published and available.
```

_automate/publish.sh
```diff
@@ -21,54 +32,80 @@ cargo clean
...
+	echo "Waiting for $NAME@$VERSION to be available..."
```
dvdplm, line 73:
```suggestion
	echo "Waiting for $NAME@$VERSION to become available at the registry..."
```

_automate/publish.sh
```diff
@@ -21,54 +32,80 @@ cargo clean
...
+	while : ; do
+		sleep 3
+		remote_version
+		if [ "$REMOTE_VERSION" = "$VERSION" ]; then
+			echo "🥳 $NAME@$VERSION published succesfuly."
```
dvdplm, line 78:
```suggestion
			echo "🥳 $NAME@$VERSION published succesfully."
```

_automate/publish.sh
```diff
@@ -21,54 +32,80 @@ cargo clean
...
-	cd $crate
-	VERSION=$(grep "^version" ./Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
-	echo "Tagging $crate@$VERSION"
+for CRATE_DIR in ${ORDER[@]}; do
+	cd $CRATE_DIR > /dev/null
+	read_toml
+	echo "Tagging $NAME@$VERSION"
 	set -x
-	git tag -a "$crate-$VERSION" -m "$crate $VERSION" || true
+	git tag -a "$NAME-$VERSION" -m "$NAME $VERSION" || true
```
dvdplm, line 94: Do you think we should fetch all remote tags first, `git fetch --tags`, to make sure we don't end up with a mismatch between local and remote tags?

_automate/publish.sh
```diff
@@ -21,54 +32,80 @@ cargo clean
...
-
+sleep 3
 git push --tags
+set +x
 
-VERSION=$(grep "^version" ./core/Cargo.toml | sed -e 's/.*"\(.*\)"/\1/')
+cd core > /dev/null
+read_toml
+cd - > /dev/null
 echo "Tagging main $VERSION"
```
dvdplm, line 107: "main"?

_automate/publish.sh
```diff
@@ -21,54 +32,80 @@ cargo clean
...
+set -x
 git tag -a v$VERSION -m "Version $VERSION"
```
dvdplm, line 109: What's the point of having an annotated tag? Doesn't seem to add much value?

_automate/publish.sh
tomusdrw: Well, we are limited to our `jsonrpc-` prefixed stuff, so I think it's good enough for our use case:)

_automate/publish.sh
tomusdrw: Makes sense! Will add.

_automate/publish.sh
tomusdrw: Since historically we release the entire "jsonrpc" suite together this is what I mean by "main".

_automate/publish.sh
tomusdrw: It shows up as release on Github I think. So it's easier to just edit it to add release notes.

_automate/publish.sh
dvdplm: I was thinking ahead about adopting this in `parity-common` but maybe I should wait until it reaches the AI Singularity or at least Turing completness... ;)

_automate/publish.sh
dvdplm: Got it. I'd drop it tbh, but not important.

_automate/publish.sh
dvdplm: ah, that's nice. TIL.

_automate/publish.sh
tomusdrw: Changed to `Tagging jsonrpc@$VERSION`

Issue #414, opened 2019-04-02
Improve client code generation


tomusdrw:
See https://github.com/paritytech/jsonrpc/pull/405#issuecomment-478912434

Issue #485, opened 2019-08-29
Switch to std::future::Future


ebkalderon:
Transitioning from `futures` 0.1 to `std::future::Future` would greatly improve the user experience in the near future, especially with async/await stabilizing in Rust 1.39 and `std::future::Future` compatible versions of `tokio`, `hyper`, and `tower` on the horizon. It seems that `jsonrpc-core` is currently the only library in my personal JSON-RPC projects that still relies exclusively on `futures` 0.1, and it is blocking me from switching over to the new trait completely. I understand if this issue will sit on the backburner for the time being until at least `tokio` 0.2.0 is released, though.

tomusdrw:
Yeah, I think it would be great to switch at least `jsonrpc-core` (and derive maybe) initially, the transports would be fine with the `Compat` layer.

@tomaka you have some experience with porting to `std::future`, any hints? :)

tomaka:
Most of the transition should be straight-forward, except that you might have to add `where F: Unpin` a bit everywhere.

One part where it might be difficult is that libraries compatible with new futures have ditched lots of combinators in favour of async-await-friendly methods that borrow objects. For example, calling `incoming()` of `async-std::TcpListener` (which returns a stream of incoming connections) no longer takes ownership of the `TcpListener` but borrows it. This is very use-friendly with async/await, but is a pain in the ass if you don't have access to async/await yet.

Also, unless that changed, if you use tokio for networking, you will have to use tokio for executing futures as well. I'd encourage using `futures-timer` and `async-std` instead of `tokio`.

ebkalderon:
According to [Rust Forge](https://forge.rust-lang.org), version 1.39 will land on November 7th, which will include stable async/await. In that case, should we start transitioning to `std::future::Future` now? Or should we wait for the API redesign that @tomaka is working on to land first?

tomusdrw:
I think it's worth start the transition, I didn't have time to look into it yet. I'd be really happy to accept a PR though if you are interested in working on this @ebkalderon :)

ebkalderon:
@tomusdrw Sure, I suppose I can try!

ebkalderon:
I've successfully converted `jsonrpc-core` to `std::future::Future`, but I am blocked on `jsonrpc-server-utils` since it requires `tokio`, and `tokio-sync >=0.2.0-alpha.2` apparently requires a nightly compiler and async/await.

tomusdrw:
[...]

Issue #422, opened 2019-04-19
Server should be an executable Future, not an executor (shim) in its own right


vorot93:
jsonrpc's ServerBuilder accepts a custom executor or starts its own and then requires the user to `start` it.

On the other hand, both [hyper](https://docs.rs/hyper) and [tarpc](https://docs.rs/tarpc) build a Server struct that is a plain `Future` that can be spawned on user's executor of choice. It is way more ergonomic and in line with async ecosystem's move towards separation of logic and its execution.

tomusdrw:
@vorot93 I agree 100%. Old API is more of a legacy from the times where `tokio/futures` and `async` in general was not really a thing yet, we should think on rewriting this if all transports support async.

christobias:
With a cursory look through the current code, IMHO since ServerHandler is a valid hyper `Service`, to support this without a breaking API change it would be a good idea to either:

1) Allow taking the `ServerHandler` from an existing `Server`, or
2) Split the creation of ServerHandler from `Server::serve_http()` and provide a way to create `Server`
 instances from existing `ServerHandler`s

The latter option would be a much better fit because users who require just the handler (because they have an existing tokio runtime or hyper instance) can utilize their existing code without extra overhead, while giving a simpler ready-made instance for easier setup

vorot93:
Actually, this may no longer be a valid issue, given that [hyper spawns tasks itself](https://docs.rs/hyper/0.13.7/hyper/server/struct.Builder.html#method.executor) now.

Where I'm contributing (e.g. [rust-devp2p](https://github.com/rust-ethereum/devp2p)) I'm leaning towards such hybrid approach:
```
impl Server {
    fn new(executor) -> Self
}

impl Future for Server {
    ....
}
```

Server itself may spawn tasks e.g. to handle each request, while providing a future that resolves when the server is shut down.

PR #565, opened 2020-07-22
Migrate core to futures=0.3

tomusdrw:
This is the first step towards #485 (also see #422). The PR rewrites `jsonrpc-core` & `jsonrpc-derive` & `jsonrpc-core-client` (but not transports) to be compatible with `futures=0.3`.

`jsonrpc-derive` is limited to return either `Result<_>` or `BoxFuture<Result<_>>` (see `WrapFuture` trait), supporting custom future types would require a bit more work. Actually it would be best to check out if we can integrate it with `async_trait`, so that the implementations can simply define `async` functions.

The transports (servers) currently run the futures in `compat` mode, they should be updated as well.

### DIFF
--- a/core-client/transports/src/lib.rs
+++ b/core-client/transports/src/lib.rs
@@ -2,15 +2,19 @@
 
 #![deny(missing_docs)]
 
-use std::pin::Pin;
 use failure::{format_err, Fail};
-use jsonrpc_core::{Error, Params};
 use jsonrpc_core::futures::channel::{mpsc, oneshot};
-use jsonrpc_core::futures::{self, future, Future, Stream, StreamExt, TryFutureExt, task::{Context, Poll}};
-use serde::Serialize;
+use jsonrpc_core::futures::{
+	self, future,
+	task::{Context, Poll},
+	Future, Stream, StreamExt, TryFutureExt,
+};
+use jsonrpc_core::{Error, Params};
 use serde::de::DeserializeOwned;
+use serde::Serialize;
 use serde_json::Value;
 use std::marker::PhantomData;
+use std::pin::Pin;
 
 pub mod transports;
 
@@ -115,10 +119,7 @@ impl From<SubscribeMessage> for RpcMessage {
 pub struct RpcChannel(mpsc::UnboundedSender<RpcMessage>);
 
 impl RpcChannel {
-	fn send(
-		&self,
-		msg: RpcMessage,
-	) -> Result<(), mpsc::TrySendError<RpcMessage>> {
+	fn send(&self, msg: RpcMessage) -> Result<(), mpsc::TrySendError<RpcMessage>> {
 		self.0.unbounded_send(msg)
 	}
 }
@@ -156,18 +157,17 @@ impl<T> TypedSubscriptionStream<T> {
 impl<T: DeserializeOwned + Unpin + 'static> Stream for TypedSubscriptionStream<T> {
 	type Item = RpcResult<T>;
 
-    fn poll_next(
-        mut self: Pin<&mut Self>,
-        cx: &mut Context,
-    ) -> Poll<Option<Self::Item>> {
+	fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
 		let result = futures::ready!(self.stream.poll_next_unpin(cx));
 		match result {
-			Some(Ok(value)) =>
-				Some(serde_json::from_value::<T>(value)
-					.map_err(|error| RpcError::ParseError(self.returns.into(), error.into()))),
+			Some(Ok(value)) => Some(
+				serde_json::from_value::<T>(value)
+					.map_err(|error| RpcError::ParseError(self.returns.into(), error.into())),
+			),
 			None => None,
 			Some(Err(err)) => Some(Err(err.into())),
-		}.into()
+		}
+		.into()
 	}
 }
 
@@ -191,10 +191,12 @@ impl RawClient {
 			sender,
 		};
 		match self.0.send(msg.into()) {
-			Ok(()) => future::Either::Left(async move { match receiver.await {
-				Ok(v) => v,
-				Err(e) => Err(RpcError::Other(e.into())),
-			}}),
+			Ok(()) => future::Either::Left(async move {
+				match receiver.await {
+					Ok(v) => v,
+					Err(e) => Err(RpcError::Other(e.into())),
+				}
+			}),
 			Err(error) => future::Either::Right(future::ready(Err(RpcError::Other(error.into())))),
 		}
 	}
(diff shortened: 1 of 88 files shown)


core/src/calls.rs
```diff
@@ -59,14 +83,24 @@ impl<T: Metadata> fmt::Debug for RemoteProcedure<T> {
 	}
 }
 
-impl<F: Send + Sync + 'static, X: Send + 'static, I> RpcMethodSimple for F
+// TODO [ToDr] Check all the bounds
```
ascjones, line 86: Still TODO?

core/src/calls.rs
```diff
@@ -11,18 +11,42 @@ impl<T: Metadata> Metadata for Option<T> {}
 impl<T: Metadata> Metadata for Box<T> {}
 impl<T: Sync + Send + 'static> Metadata for Arc<T> {}
 
+/// A future-conversion trait.
+pub trait WrapFuture<T, E> {
```
ascjones, line 15: Why not just call it `IntoFuture`? To avoid confusion with the erstwhile `futures` trait?

core/src/io.rs
```diff
@@ -260,8 +273,8 @@ impl<T: Metadata, S: Middleware<T>> MetaIoHandler<T, S> {
 				let valid_version = self.compatibility.is_version_valid(jsonrpc);
 
 				let call_method = |method: &Arc<dyn RpcMethod<T>>| {
-					let method = method.clone();
-					futures::lazy(move || method.call(params, meta))
+					// TODO [ToDr] lazy here?
```
ascjones, line 276: I don't think it's necessary AFAICT

derive/src/rpc_trait.rs
```diff
@@ -258,8 +258,10 @@ pub fn rpc_impl(input: syn::Item, options: &DeriveOptions) -> Result<proc_macro2
 	if options.enable_client {
 		let rpc_client_module = generate_client_module(&method_registrations, &rpc_trait, options)?;
 		submodules.push(rpc_client_module);
+		let client_name = syn::Ident::new(&format!("{}Client", name), proc_macro2::Span::call_site());
 		exports.push(quote! {
 			pub use self::#mod_name_ident::gen_client;
+			pub use self::#mod_name_ident::gen_client::Client as #client_name;
```
ascjones, line 264: Nice. Probably closes https://github.com/paritytech/jsonrpc/issues/414?

pubsub/src/subscription.rs
```diff
@@ -290,19 +294,25 @@ where
 				let unsub = self.unsubscribe.clone();
 				let notification = self.notification.clone();
 				let subscribe_future = rx.map_err(|_| subscription_rejected()).and_then(move |result| {
-					futures::done(match result {
+					futures::future::ready(match result {
 						Ok(id) => {
 							session.add_subscription(&notification, &id, move |id| {
-								let _ = unsub.call(id, None).wait();
+								// TODO [ToDr] We currently run unsubscribe tasks on a shared thread pool.
```
ascjones, line 300: Might be worth adding a follow up issue for this otherwise this TODO will be here forever

derive/Cargo.toml
```diff
@@ -19,12 +19,12 @@ quote = "1.0.6"
...
-futures = "~0.1.6"
+assert_matches = "1.3"
+jsonrpc-core = { version = "15.0", path = "../core" }
+jsonrpc-core-client = { version = "15.0", path = "../core-client" }
+jsonrpc-pubsub = { version = "15.0", path = "../pubsub" }
+jsonrpc-tcp-server = { version = "15.0", path = "../tcp" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-tokio = "0.1"
+tokio = "0.2"
```
ascjones, line 29: I don't think this is required anymore

ws/src/metadata.rs
```diff
@@ -78,10 +78,12 @@ pub struct RequestContext {
 impl RequestContext {
 	/// Get this session as a `Sink` spawning a new future
 	/// in the underlying event loop.
-	pub fn sender(&self) -> mpsc::Sender<String> {
+	pub fn sender(&self) -> mpsc::UnboundedSender<String> {
+		use futures03::{StreamExt, TryStreamExt};
 		let out = self.out.clone();
-		let (sender, receiver) = mpsc::channel(1);
-		self.executor.spawn(SenderFuture(out, receiver));
+		let (sender, receiver) = mpsc::unbounded();
```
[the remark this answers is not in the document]
tomusdrw: The channel was used incorrectly, since it was `cloned` every time before issuing `send`. The spec says that `1` is a reserved value, but there is also at least 1 additional entry for every clone of `Sender`. So basically the channel could never really be full, since each `send` was extending it anyway. I think it could be set to `0` here, but possible it panics, so it's just set to the minimal possible value.

tcp/src/dispatch.rs
```diff
@@ -3,24 +3,25 @@ use std::collections::HashMap;
 use std::net::SocketAddr;
 use std::sync::Arc;
 
-use crate::jsonrpc::futures::sync::mpsc;
-use crate::jsonrpc::futures::{Async, Future, Poll, Sink, Stream};
+use crate::jsonrpc::futures::{self as futures03, channel::mpsc, StreamExt};
```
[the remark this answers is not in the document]
tomusdrw: I'd prefer to use the re-exported one to avoid rewriting the code in the future. If `tcp` get's rid of compat requirements we can simply remove the `futures03` dependency from Cargo.toml. Added a todo for this.

pubsub/src/subscription.rs
tomusdrw: Done, see #570

core/src/calls.rs
tomusdrw: Yeah, I want to avoid confusion with both `futures01::IntoFuture` trait and the `futures03::IntoFuture` struct. I think long-term we should consider `jsonrpc-derive` to require the methods to be `async` and then this issue is no longer a thing.

tomusdrw: Added an issue: https://github.com/paritytech/jsonrpc/issues/571

core-client/transports/src/lib.rs
```diff
@@ -225,29 +183,34 @@ impl From<RpcChannel> for RawClient {
...
-			.and_then(|_| RpcFuture::new(receiver))
+		match self.0.send(msg.into()) {
+			Ok(()) => future::Either::Left(async move {
+				match receiver.await {
+					Ok(v) => v,
+					Err(e) => Err(RpcError::Other(e.into())),
+				}
+			}),
+			Err(error) => future::Either::Right(future::ready(Err(RpcError::Other(error.into())))),
+		}
```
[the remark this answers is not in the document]
tomusdrw: I did try making the whole method `async`, but the issue is with the lifetimes. The required future must be `'static`, and afaict there is no syntax to express that if you have `async fn`. So instead I rewrote the part that is really "async" into an `async move` block, but the outer function still has `'static` bound to prevent issues.

tomusdrw:
@ascjones I believe I've addressed all of your comments now, would you mind taking another look?

PR #572, opened 2020-07-29
Fix local transport wakeups.

tomusdrw:


tomusdrw:
Fix working confirmed by @seunlanlege

PR #573, opened 2020-07-29
Make sure manager is Sync+Send

tomusdrw:


tomusdrw:
This prevents using `LocalSpawner`, but in practice it wouldn't work anyway, because jsonrpc-derive trait implementations enforce `Send + Sync`.

PR #584, opened 2020-09-17
Use patched ws-rs

maciejhirsz:
Revert to `ws-rs` once https://github.com/housleyjk/ws-rs/pull/328 is merged and published.


tomusdrw (approved):

dvdplm (approved):

ordian (approved):
Just to be clear, we probably don't want to publish v16 from master yet, as this change will likely get reverted when the fix is merged and published upstream.

Looks good to me.

PR #587, opened 2020-09-21
Fix typos

nazar-pc:


dvdplm (commented):
Thank you for the contribution.

[`AcqRel`](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.AcqRel) is a valid `Ordering`, can you elaborate on why `SeqCst` is better here?

nazar-pc:
It is not better, `AcqRel` is not valid for both for load and store operations. You will get panic "there is no such thing as an acquire/release load" in runtime.
It can be used for things like swap or compare_and_swap where you combine load and store in one call.
`SeqCst` is just a safe default, especially for examples.

niklasad1 (approved):
LGTM, thank you.

dvdplm (approved):

PR #588, opened 2020-09-23
Update examples in top-level readme.

### LINKED ISSUE #586: Derive pubsub example in docs is not working
I've copy-pasted the example given (even though it is not complete) and got this compile-time error:
```
error[E0271]: type mismatch resolving `<jsonrpc_pubsub::IoDelegate<sub::RpcImpl, Arc<Session>> as IntoIterator>::Item == (std::string::String, RemoteProcedure<()>)`
  --> src/rpc.rs:95:8
   |
95 |     io.extend_with(sub::RpcImpl::default().to_delegate());
   |        ^^^^^^^^^^^ expected struct `Arc`, found `()`
   |
   = note: expected tuple `(std::string::String, RemoteProcedure<Arc<Session>>)`
              found tuple `(std::string::String, RemoteProcedure<()>)`
```

I'm using current git master because I don't want to mess with futures 0.1.

Any ideas what is wrong here? Also having a full example in there would help to make sure it doesn't regress.

UPD: Seems to be caused by `IoHandler::new();` instead of `PubSubHandler::new(MetaIoHandler::default())`, but now a new error:
```
error[E0277]: the trait bound `Session: Default` is not satisfied
  --> src/rpc.rs:82:37
   |
82 |     let server = ServerBuilder::new(io)
   |                                     ^^ the trait `Default` is not implemented for `Session`
   | 
  ::: /home/nazar-pc/.cargo/git/checkouts/jsonrpc-22a65938f7d26b63/92e3c14/ws/src/server_builder.rs:37:12
   |
37 |         T: Into<core::MetaIoHandler<M, S>>,
   |            ------------------------------- required by this bound in `ServerBuilder::<M, S>::new`
   |
   = note: required because of the requirements on the impl of `Default` for `Arc<Session>`
```

UPD 2: `pubsub/examples/pubsub.rs` is also broken, starting with `context.sender.clone()` that needs to be `context.sender()` to `there is no such thing as an acquire/release load` for obvious reasons; did it ever work at all?

tomusdrw:
Closes #586


derive/src/lib.rs
```diff
@@ -120,7 +120,20 @@
...
+//!
+//!	  let server_builder = jsonrpc_tcp_server::ServerBuilder::with_meta_extractor(
+//!		io,
+//!		|request: &jsonrpc_tcp_server::RequestContext| Arc::new(Session::new(request.sender.clone()))
+//!	  );
+//!	  let server = server_builder
+//!		.start(&"127.0.0.1:3030".parse().unwrap())
+//!		.expect("Unable to start TCP server");
+//!
+//!	  // server.wait();
```
niklasad1, line 135: "uncommented" because the example wouldn't terminate otherwise?

niklasad1 (approved):

derive/src/lib.rs
tomusdrw: Yes, is there some other way to prevent that? I guess we could mark it as "no-run"?

niklasad1: I guess we could either mark it `no_run` or hide the "uncommented" line from the docs by `# // server.wait();`

tomusdrw: 1. `no_run` feels inferior, since it won't detect the runtime errors
2. I think hidden lines are still being executed (they are just not displayed in the example)

I can remove the comment completely, but I think it's easier if we make users aware they have to wait for the server. Can add an explanation comment.

dvdplm: > Can add an explanation comment.

Yes please. :)

tomusdrw: Added.

niklasad1:
//cc @dvdplm @maciejhirsz 

need 1 more review :)

dvdplm (approved):

PR #589, opened 2020-09-25
Update `local` client to support middleware (Kudos Seun)

tomusdrw:
Previously the `jsonrpc_core_client::transports::local::connect` consumed a `Deref<Target=MetaIoHandler<TMetadata, NoopMiddleware>>`,

This PR corrects that behaviour allowing the local transport support other `Middleware` implementations

### DIFF
--- a/core-client/transports/src/transports/local.rs
+++ b/core-client/transports/src/transports/local.rs
@@ -85,7 +85,7 @@ where
 }
 
 /// Connects to a `Deref<Target = MetaIoHandler<Metadata>`.
-pub fn connect_with_metadata<TClient, THandler, TMetadata, TMiddleware>(
+pub fn connect_with_metadata_and_middleware<TClient, THandler, TMetadata, TMiddleware>(
 	handler: THandler,
 	meta: TMetadata,
 ) -> (TClient, impl Future<Item = (), Error = RpcError>)
@@ -101,8 +101,21 @@ where
 	(client, rpc_client)
 }
 
+/// Connects to a `Deref<Target = MetaIoHandler<Metadata>`.
+pub fn connect_with_metadata<TClient, THandler, TMetadata>(
+	handler: THandler,
+	meta: TMetadata,
+) -> (TClient, impl Future<Item = (), Error = RpcError>)
+	where
+		TClient: From<RpcChannel>,
+		THandler: Deref<Target = MetaIoHandler<TMetadata>> + Unpin,
+		TMetadata: Metadata,
+{
+	connect_with_metadata_and_middleware(handler, meta)
+}
+
 /// Connects to a `Deref<Target = MetaIoHandler<Metadata + Default>`.
-pub fn connect<TClient, THandler, TMetadata, TMiddleware>(
+pub fn connect_with_middleware<TClient, THandler, TMetadata, TMiddleware>(
 	handler: THandler,
 ) -> (TClient, impl Future<Item = (), Error = RpcError>)
 where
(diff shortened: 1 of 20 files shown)


dvdplm (commented):
A PR description would be good.

This adds a new type parameter; isn't that a breaking change?

seunlanlege (approved):

core-client/transports/src/transports/local.rs
```diff
@@ -98,10 +102,11 @@ where
 }
 
 /// Connects to a `Deref<Target = MetaIoHandler<Metadata + Default>`.
-pub fn connect<TClient, THandler, TMetadata>(handler: THandler) -> (TClient, impl Future<Item = (), Error = RpcError>)
+pub fn connect<TClient, THandler, TMetadata, TMiddleware>(handler: THandler) -> (TClient, impl Future<Item = (), Error = RpcError>)
```
tomusdrw, line 105: As @dvdplm noticed it's a breaking change (so would require bumping to `16`), perhaps better to add a new set of methods like `connect_with_middleware` or set the default type somehow?
Usage of middlewares is not super common, so I think it's better to keep the default `connect` stuff simple.

niklasad1 (commented):
Is it reasonable to merge with master instead after the breaking changes? I think there is some unreleased stuff there (futures v0.3) IIRC.

seunlanlege:
any luck this can be merged and released today?

tomusdrw:
@seunlanlege Please see https://github.com/paritytech/jsonrpc/pull/589#discussion_r494942557
This is a breaking change it either requires bump to 16.0 or we must introduce it in a backward compatible way.

seunlanlege:
Weird I thought it could be released under 15.1? Either way, bumping to 16 feels extreme. Adding a new method, `connect_with_middleware` feels mroe optimal

core-client/transports/src/transports/local.rs
```diff
@@ -17,10 +17,11 @@ pub struct LocalRpc<THandler, TMetadata> {
 	receiver: mpsc::UnboundedReceiver<String>,
 }
 
-impl<TMetadata, THandler> LocalRpc<THandler, TMetadata>
+impl<TMetadata, THandler, TMiddleware> LocalRpc<THandler, TMetadata>
 where
 	TMetadata: Metadata,
-	THandler: Deref<Target = MetaIoHandler<TMetadata>>,
+	TMiddleware: Middleware<TMetadata>,
+	THandler: Deref<Target = MetaIoHandler<TMetadata, TMiddleware>> + Unpin,
```
niklasad1, line 24: Can you explain why `Unpin` is required here? It worked without it before this PR and I don't understand why is it needed now.

Technically, a breaking change too but I guess it's used mostly by `IoHandler`.

core-client/transports/src/transports/local.rs
```diff
@@ -97,24 +101,55 @@ where
...
 	TMetadata: Metadata + Default,
 {
-	connect_with_metadata(handler, Default::default())
+	connect_with_metadata_and_middleware(handler, Default::default())
+}
+
+/// Connects to a `Deref<Target = MetaIoHandler<Metadata + Default>`.
+pub fn connect<TClient, THandler, TMetadata>(
```
niklasad1, line 131: hrm, this will connect without `Middleware (Noop)` because of `MetaIoHandler<TMetadata>`? It seems unintuitive to me but I understand that it wouldn't be backward compatible otherwise. Perhaps explain this with better documentation and remove it in `v16` later?

seunlanlege: it will connect with the `NoopMiddleware`

niklasad1 (commented):
@seunlanlege 

The repo is using `cargo fmt` please use it and the CI will pass.

core-client/transports/src/transports/local.rs
niklasad1: Yepp, that was what I meant with `Middleware (Noop)`, but I was mostly referring to:

> Connects to a `Deref<Target = MetaIoHandler<Metadata + Default>`

It's not easy to understand that is using the default associated type `NoopMiddleware` at least not to me.

niklasad1 (dismissed):

niklasad1 (changes requested):
The PR is now mixing versions `15.1.0` and `15.0.1` which causes the build to fail.

IMHO, we should bump `minor` because it actually adds functionality in backward-compatible manner. If you don't agree please convince me :) 

EDIT:
use `./_automate/bump_version.sh` to bump versions

niklasad1 (approved):

tomusdrw (commented):
lgtm!

maciejhirsz (approved):

PR #594, opened 2020-10-01
Expose direct configuration of max in and out buffer capacity

### LINKED ISSUE #590: jsonrpc_ws_server::ServerBuilder doesn't allow configuring separate values for buffer capacity
#### Problem
As of v15, the `jsonrpc_ws_server::ServerBuilder::max_payload` value is used for max frame size and both incoming and outgoing buffer capacities making it impossible to configure these buffer values separately. A use case could be a server which wishes to limit request size to a small size but respond with relatively large responses.

#### Proposed Changes
Add methods for setting max buffer sizes:
```
fn max_in_buffer_size()
fn max_out_buffer_size()
```

emostov:
Closes #590, see issue for feature description.

If not set, `max_in_buffer_capacity` and `max_out_buffer_capacity` explicitly defaults to 10,485,760, which is the same as [`parity-ws` `Settings::default`](https://github.com/paritytech/ws-rs/blob/be4d47575bae55c60d9f51b47480d355492a94fc/src/lib.rs#L254).

### DIFF
--- a/ws/src/server_builder.rs
+++ b/ws/src/server_builder.rs
@@ -148,7 +148,7 @@ where
 			self.max_connections,
 			self.max_payload_bytes,
 			self.max_in_buffer_capacity,
-			self.max_in_buffer_capacity,
+			self.max_out_buffer_capacity,
 		)
 	}
 }
--- a/ws/src/server_builder.rs
+++ b/ws/src/server_builder.rs
@@ -174,7 +174,7 @@ mod tests {
 	#[test]
 	fn config_usize_vals_can_be_set() {
 		let server = basic_server_builder();
-		
+
 		// We can set them individually
 		let server = server.max_connections(10);
 		assert_eq!(server.max_connections, 10);
--- a/ws/src/server_builder.rs
+++ b/ws/src/server_builder.rs
@@ -24,6 +24,8 @@ pub struct ServerBuilder<M: core::Metadata, S: core::Middleware<M>> {
 	executor: UninitializedExecutor,
 	max_connections: usize,
 	max_payload_bytes: usize,
+	max_in_buffer_capacity: usize,
+	max_out_buffer_capacity: usize,
 }
 
 impl<M: core::Metadata + Default, S: core::Middleware<M>> ServerBuilder<M, S>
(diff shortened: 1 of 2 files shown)


dvdplm (changes requested):
Looks good, let's add a test?

ws/src/server_builder.rs
```diff
@@ -129,6 +147,8 @@ where
 			self.executor,
 			self.max_connections,
 			self.max_payload_bytes,
+			self.max_in_buffer_capacity,
+			self.max_in_buffer_capacity,
```
niklasad1, line 151:
```suggestion
			self.max_out_buffer_capacity,
```

emostov:
@niklasad1, @maciejhirsz, @dvdplm - added some basic tests. I originally was going to assert every config value, but I could not think of a zero overhead way around the fact some of the more complex types don't support equality comparisons. It also looks like those complex types are tested implicitly in some of the higher level tests so I think that means technically they are covered. 

Unrelated to the tests, I realized this PR may break the API since it changes the defaults for `max_in_buffer_capacity` and `max_out_buffer_capacity`. Previously they defaulted to 5,242,880 (the default of `max_payload_bytes`), now they default to `10,485,760`.

niklasad1:
IMO, the tests are sufficient for this change

> Previously they defaulted to 5,242,880 (the default of max_payload_bytes), now they default to 10,485,760

Ok, basically because `ws-rs` defaults `max_buffer_i/o_size` 10MB it's used here too but configured explicitly, otherwise please explain?!

If we want to go with 10MB instead it would be reasonable to use it for `max_payload_bytes` too?

However, I'm not familiar with the reasoning about these limits but 5MB should cover most use-cases AFAIK but perhaps @maciejhirsz or @tomusdrw could explain the reasoning about these limits :)

tomusdrw:
> However, I'm not familiar with the reasoning about these limits but 5MB should cover most use-cases AFAIK but perhaps @maciejhirsz or @tomusdrw could explain the reasoning about these limits :)

I can't remember exactly, but wasn't this the same limit as `ws-rs` was setting? If not then the values are probably arbitrary, fine to change them to whatever, obv with semver compliance.

maciejhirsz (approved):

maciejhirsz:
For the record, the default 10mb limit in `ws-rs`/`parity-ws` is there just so that we don't have an unlimited default. I didn't do any science for it, it should be high enough to make everyday regular use work without issues, so I figured it's better to err on the larger size rather than smaller as long as it doesn't crash things.

emostov:
Just need one more approval :)

niklasad1:
@dvdplm: can you take a look at this again?

dvdplm (approved):

PR #597, opened 2020-11-13
Alter headers

enfipy:
This PR alters headers `ACCEPT` and content type for `text/json`.

The ideal solution is to provide a method to change it via lib API but as a workaround - this will help https://github.com/adoriasoft/polkadot_cosmos_integration.

### DIFF
--- a/http/src/handler.rs
+++ b/http/src/handler.rs
@@ -544,6 +544,7 @@ where
 		if is_options {
 			headers.append(header::ALLOW, allowed.clone());
 			headers.append(header::ACCEPT, HeaderValue::from_static("application/json"));
+			headers.append(header::ACCEPT, HeaderValue::from_static("text/json"));
 		}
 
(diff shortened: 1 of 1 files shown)


tomusdrw:
Can't find `text/json` in supported MIME types: https://www.iana.org/assignments/media-types/media-types.xhtml#text

I'm reluctant to add that, since it may have security implications (`application/json` requires pre-flight request for CORS in browsers, I'm not sure if that's going to be the case for `text/json` as well).

I feel this is more of a problem of the client library you are using than the server, similarly how we don't accept malformed HTTP requests I'm not yet convinced we should accept malformed mime types.

enfipy:
You are right. So I think it's better to update tendermint instead: https://github.com/tendermint/tendermint/blob/346aa14db560c3c867a3e00b4fbd8b000dcffe30/rpc/jsonrpc/client/http_json_client.go#L219

PR #598, opened 2020-11-18
chore(deps): update parking_lot 0.11.1

niklasad1:
Fixes https://github.com/RustSec/advisory-db/pull/483/files


niklasad1:
I was wrong `lock_api 0.4.2` is not included in this release, thus one would have to do `cargo update -p lock_api` but because we don't commit `Cargo.lock` it doesn't make sense anymore.

Let's wait for the next release.

Issue #607, opened 2021-01-15
Cut a new release with Tokio 0.2 support


Xanewok:
Now that #603 is merged, it'd be good to cut a major release containing it. Because the Tokio ecosystem is currently somewhat split between Tokio 0.1, 0.2, and 0.3/1.0, I believe it'd be beneficial to cut a release for projects that are still stuck in the 0.2; either because it's a non-trivial upgrade or because their dependencies still rely on the 0.2 version.

Once we release that, we can safely upgrade to Tokio 1.0 and get rid of some old deps, e.g. #575.

tomusdrw:
v17.0 is out.

Issue #605, opened 2021-01-04
Migrate to soketto for Websocket dependency


Xanewok:
Right now we use (unmaintained) `websocket` dependency that pulls older versions of hyper (0.10) and tokio (0.1). The `soketto` dependency is transport-agnostic and so should help reduce the dependency bloat.

Blocked by https://github.com/paritytech/jsonrpc/pull/603 (migration to `futures v0.3`).

Xanewok:
Addressing this will probably also close #158.

PR #608, opened 2021-01-19
Bump version.

tomusdrw:
Related #607


niklasad1 (approved):
LGTM

dvdplm (approved):

Xanewok (commented):
Looks good, thanks! Do we want to address https://github.com/paritytech/jsonrpc/issues/605 as well before cutting the new release?

tomusdrw:
> Looks good, thanks! Do we want to address #605 as well before cutting the new release?

@Xanewok your call. IMHO better to do it separately, and see if we run into any regressions without soketto, perhaps #605 could be implemented without an API change so it would be transparent to the users.

Xanewok:
@tomusdrw Makes sense, let's do that separately. Deploy away! :rocket:

Xanewok (approved):

PR #609, opened 2021-01-20
core: Allow for using only futures-util dependency

### LINKED ISSUE #604: core should depend on a subset of futures
It's kind of annoying that it pulls in the entire `futures` stack (futures-executor, etc.), since it's just a bunch of type definitions. `futures-core` should be enough.

Xanewok:
Fixes #604

This makes `cargo build --no-default-features` work by pulling only `futures-util` (and transitively `futures-task`). It's worth noting that:
1. the convenience `handle_request_sync` function is only available when using `futures-executor` feature,
2. ...and so `cargo test` only works with default feature set (no idea how to cleanly enable the crate features for unit tests).

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -29,7 +29,7 @@ extern crate serde_derive;
 
 #[cfg(feature = "futures-executor")]
 pub use futures_executor;
-pub use futures;
+pub use futures_util;
 
 #[doc(hidden)]
 pub extern crate serde;
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -27,6 +27,8 @@ extern crate log;
 #[macro_use]
 extern crate serde_derive;
 
+#[cfg(feature = "futures")]
+pub use futures;
 #[cfg(feature = "futures-executor")]
 pub use futures_executor;
 pub use futures_util;
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -45,7 +45,7 @@ pub mod types;
 pub type Result<T> = std::result::Result<T, Error>;
 
 /// A `Future` trait object.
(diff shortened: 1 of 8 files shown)


tomusdrw (approved):
Initially `futures` was re-exported to make it less annoying to import `futures` library just for `Future` trait. Now with async/await and `std::future::Future` we should get rid of this long-term.

Xanewok:
`gitlab-test-windows-stable` failure seems to be spurious:
```
$ cargo build --all
error: failed to run `rustc` to learn about target-specific information
Caused by:
  process didn't exit successfully: `sccache rustc - --crate-name ___ --print=file-names --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit code: 2)
  ```
  but I can't seem to be able to re-run it :cry:

**Comment by tomusdrw** [comment]
@Xanewok restarted.

File: core/src/lib.rs [hunk 1]
```diff
@@ -45,7 +49,7 @@ pub mod types;
 pub type Result<T> = std::result::Result<T, Error>;
 
 /// A `Future` trait object.
-pub type BoxFuture<T> = Pin<Box<dyn futures::Future<Output = T> + Send>>;
+pub type BoxFuture<T> = Pin<Box<dyn std::future::Future<Output = T> + Send>>;
```
niklasad1, line 52: any reason not to use https://docs.rs/futures/0.3.13/futures/future/type.BoxFuture.html instead of this?

niklasad1 (approved):
LGTM,

> ...and so cargo test only works with default feature set (no idea how to cleanly enable the crate features for unit tests).

https://github.com/rust-lang/cargo/issues/2911 :grimacing:

PR #619, opened 2021-03-09
Return error when unsubscribing from invalid subscription id

jstarry:
#### Problem
When processing an unsubscribe request, the unsubscribe callback is called even if there was no active subscription for that id.

#### Changes
- Return an error for invalid unsubscribe requests
- Don't call unsubscribe callback for invalid unsubscribes

cc @tomusdrw

### DIFF
--- a/pubsub/Cargo.toml
+++ b/pubsub/Cargo.toml
@@ -21,6 +21,7 @@ serde = "1.0"
 
 [dev-dependencies]
 jsonrpc-tcp-server = { version = "17.0", path = "../tcp" }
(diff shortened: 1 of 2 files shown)


tomusdrw (approved):
nice catch, thanks!

jstarry:
No problem, thanks for the quick review! Do you think this could get released today? I'd like to get the fix in asap, but can build off a fork if needed. Cheers!

jstarry:
Hey @tomusdrw! Just a friendly ping about getting this out into a new release 😄

PR #620, opened 2021-03-16
Fix link to documentation

taiki-e:
The link to `jsonrpc-client-transports` documentation in [jsonrpc-core-client documentation](https://docs.rs/jsonrpc-core-client/17.0.0/jsonrpc_core_client/) currently points [unexist crate](https://docs.rs/jsonrpc-core-client/17.0.0/jsonrpc_client_transports/).

PR #624, opened 2021-03-31
Bump minor version.

tomusdrw:
Because of #609 

cc @niklasad1 please let me know if that's good for you to bump to `17.1`, alternatively we can just release bugfix updates for `core-transport` (#620) and `pubsub` (#619) and leave `core` for later.


niklasad1 (approved):
I'm all good with including #609 it looks quite straight forward to me.

dvdplm (approved):

archseer:
Could you also publish `core@17.1.0` as a crate?

PR #625, opened 2021-04-19
Add clippy to CI and fix several clippy warnings

chevdor:
This PR add running clippy right after rustfmt during the CI.
It also fixes many warnings (many being due to mix of spaces and tabs in the doc examples).

### DIFF
--- a/derive/src/rpc_attr.rs
+++ b/derive/src/rpc_attr.rs
@@ -71,7 +71,7 @@ impl RpcMethodAttribute {
 	fn parse_meta(attr: &syn::Attribute, output: &syn::ReturnType) -> Option<Result<RpcMethodAttribute>> {
 		match attr.parse_meta().and_then(validate_attribute_meta) {
 			Ok(ref meta) => {
-				let attr_kind = match path_to_str(meta.path()).as_ref().map(String::as_str) {
+				let attr_kind = match path_to_str(meta.path()).as_deref() {
 					Some(RPC_ATTR_NAME) => Some(Self::parse_rpc(meta, output)),
 					Some(PUB_SUB_ATTR_NAME) => Some(Self::parse_pubsub(meta)),
 					_ => None,
@@ -89,9 +89,7 @@ impl RpcMethodAttribute {
 										// "`raw_params` will be deprecated in a future release. Use `params = \"raw\" instead`"
 										Ok(Some(ParamStyle::Raw))
 									}
-									false => {
-										get_meta_list(meta).map_or(Ok(None), |ml| get_params_style(ml).map(|s| Some(s)))
-									}
+									false => get_meta_list(meta).map_or(Ok(None), |ml| get_params_style(ml).map(Some)),
 								}?;
 								Ok(RpcMethodAttribute {
 									attr: attr.clone(),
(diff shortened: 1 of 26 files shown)


derive/src/rpc_trait.rs
```diff
@@ -48,7 +48,7 @@ impl<'a> Fold for RpcTrait {
 	fn fold_trait_item_type(&mut self, ty: syn::TraitItemType) -> syn::TraitItemType {
 		if ty.ident == METADATA_TYPE {
 			self.has_metadata = true;
-			let mut ty = ty.clone();
+			let mut ty = ty;
```
tomusdrw, line 51: We could probably just do `mut ty` in the argument list.

derive/src/to_delegate.rs
```diff
@@ -244,11 +244,10 @@ impl RpcMethod {
 				quote! { let params: _jsonrpc_core::Result<_> = Ok((params,)); }
 			} else if self.attr.params_style == Some(ParamStyle::Positional) {
 				quote! { let params = params.parse::<(#(#param_types, )*)>(); }
-			} else
-			/* if self.attr.params_style == Some(ParamStyle::Named) */
-			{
+			} else {
 				unimplemented!("Server side named parameters are not implemented");
 			}
+			/* if self.attr.params_style == Some(ParamStyle::Named) */
```
tomusdrw, line 250: Looks kind of weird now, perhaps move it above `unimplemented!` call?

tomusdrw (approved):
Awesome thanks! Few tiny things to address and it's imho ready to merge.

derive/src/to_delegate.rs
chevdor: Should we not simply remove this comment ? It is 13 months old.
Do you remember the purpose of keeping it ?

tomusdrw: Nope, not really :) Yeah, let's remove it. If we wanted to be explicit about the variant that is handled here, it would be better to use `match` instead.

tomusdrw (approved):

dvdplm (approved):
TIL that `if let Err(_) = thing` is less good than `if thing.is_err()`.

derive/src/rpc_attr.rs
```diff
@@ -71,7 +71,7 @@ impl RpcMethodAttribute {
 	fn parse_meta(attr: &syn::Attribute, output: &syn::ReturnType) -> Option<Result<RpcMethodAttribute>> {
 		match attr.parse_meta().and_then(validate_attribute_meta) {
 			Ok(ref meta) => {
-				let attr_kind = match path_to_str(meta.path()).as_ref().map(String::as_str) {
+				let attr_kind = match path_to_str(meta.path()).as_deref() {
```
niklasad1, line 74: that's neat TIL

derive/src/to_client.rs
```diff
@@ -212,10 +217,9 @@ fn compute_args(method: &syn::TraitItemMethod) -> Punctuated<syn::FnArg, syn::to
 			}) => segments,
 			_ => continue,
 		};
-		let ident = match &segments[0] {
-			syn::PathSegment { ident, .. } => ident,
-		};
-		if ident.to_string() == "Self" {
+		let syn::PathSegment { ident, .. } = &segments[0];
+		let ident = ident;
```
niklasad1, line 221: why is this `let binding` necessary?

niklasad1 (approved):

PR #628, opened 2021-06-29
Update to Tokio 1.0

### LINKED ISSUE #621: Upgrade tokio to 1.x
Tokio 1.x has been released for several months ago, we need to upgrade to 1.x

Xanewok:
~Needs https://github.com/NikVolf/parity-tokio-ipc/pull/32 (pulls a `[patch.crates-io]` in the meantime).~

Closes #621

Primary motivation is to fix RLS build on aarch64 Windows (see https://github.com/rust-lang/rls/pull/1741). Tested locally on a Windows machine, hopefully unix will be happy as well.

https://tokio.rs/blog/2020-10-tokio-0-3 might be useful when reviewing

### DIFF
--- a/core-client/transports/Cargo.toml
+++ b/core-client/transports/Cargo.toml
@@ -44,11 +44,11 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 url = "1.7"
 
-hyper = { version = "0.13", optional = true }
-hyper-tls = { version = "0.4", optional = true }
+hyper = { version = "0.14", features = ["client", "http1", "server"], optional = true }
+hyper-tls = { version = "0.5", optional = true }
 jsonrpc-server-utils = { version = "17.1", path = "../../server-utils", optional = true }
 parity-tokio-ipc = { version = "0.8", optional = true }
-tokio = { version = "0.2", optional = true }
+tokio = { version = "1", optional = true }
 websocket = { version = "0.24", optional = true }
 
 [dev-dependencies]
--- a/core-client/transports/Cargo.toml
+++ b/core-client/transports/Cargo.toml
@@ -47,7 +47,7 @@ url = "1.7"
 hyper = { version = "0.14", features = ["client", "http1"], optional = true }
 hyper-tls = { version = "0.5", optional = true }
 jsonrpc-server-utils = { version = "17.1", path = "../../server-utils", optional = true }
-parity-tokio-ipc = { version = "0.8", optional = true }
+parity-tokio-ipc = { version = "0.9", optional = true }
 tokio = { version = "1", optional = true }
 websocket = { version = "0.24", optional = true }
 
--- a/core-client/transports/Cargo.toml
+++ b/core-client/transports/Cargo.toml
@@ -44,7 +44,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 url = "1.7"
 
-hyper = { version = "0.14", features = ["client", "http1", "server"], optional = true }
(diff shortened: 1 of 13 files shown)


seunlanlege (approved):

seunlanlege:
just needs to fix CI build errors

core-client/transports/Cargo.toml
```diff
@@ -44,11 +44,11 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 url = "1.7"
 
-hyper = { version = "0.13", optional = true }
-hyper-tls = { version = "0.4", optional = true }
+hyper = { version = "0.14", features = ["client", "http1", "server"], optional = true }
```
niklasad1, line 47: the server feature shouldn't be needed here right?

niklasad1 (approved):
LGTM, one question about the `server feature` for the http client

core-client/transports/Cargo.toml
Xanewok: You're right, I overlooked that - fixed in https://github.com/paritytech/jsonrpc/pull/628/commits/b05019d5cb88cccdd5a93382aeb8e4f0b1c39f8f

Xanewok:
Added some more commits since the last review. Can I get final rubber-stamp before merging this?

server-utils/src/suspendable_stream.rs
```diff
@@ -19,7 +17,7 @@ pub struct SuspendableStream<S> {
 	next_delay: Duration,
 	initial_delay: Duration,
 	max_delay: Duration,
-	timeout: Option<Delay>,
+	timeout: Option<Instant>,
```
tomusdrw, line 20: should maybe be renamed to `deadline`?

tomusdrw (approved):

server-utils/src/suspendable_stream.rs
Xanewok: Hm, `deadline` does indeed sound better but I'm having trouble understanding at a glance what this deadline is related to (it's until then that the "timeout" occurs, for lack of a better word).

Maybe something like `suspended_until` would be better in this case?

server-utils/src/suspendable_stream.rs
```diff
@@ -42,13 +42,17 @@ where
...
+			// polling from the underlying stream for a bit
+			if let Some(deadline) = &mut self.suspended_until {
+				let deadline = tokio::time::Instant::from_std(*deadline);
+				let sleep = tokio::time::sleep_until(deadline);
 				futures::pin_mut!(sleep);
 				match sleep.poll(cx) {
 					Poll::Pending => return Poll::Pending,
-					Poll::Ready(()) => {}
+					Poll::Ready(()) => {
+						self.suspended_until = None;
```
Xanewok, line 54: If we're ready then let's reset the suspension flag not to execute that code path everytime

server-utils/src/suspendable_stream.rs
Xanewok: Renamed in https://github.com/paritytech/jsonrpc/pull/628/commits/e265114ad3bdb875ff7bcae3e3c259e81b3d78e8

Xanewok:
Thanks for the reviews!

Would it be possible to release new versions of the crates somewhat soon that contain this PR?

tomusdrw:
> Would it be possible to release new versions of the crates somewhat soon that contain this PR?

I guess we should bump the version first. I guess 18.x ?

Xanewok:
Yep, sounds good. I'll prepare a bump PR if that's okay

PR #634, opened 2021-08-26
Implement server side named parameters

luc65r:


luc65r:
Oh I just found out it doesn't work with the meta argument

tomusdrw:
Good point, didn't think of that. Can we add a test for this? I guess the solution would be to either:
1. Attempt to detect the meta argument (either based on the type or the variable name)
2. Require an extra attribute/variant to skip the first parameter.

luc65r:
I fixed it, I hope I didn't change too much code.

Now it has some issues with attributes (see commented out code in `tests/macro.rs`).

derive/src/to_delegate.rs
```diff
@@ -206,21 +206,39 @@ impl RpcMethod {
...
+		for (i, arg) in args {
+			if let Some(sarg) = Self::special_arg(i, arg.clone()) {
+				special_args.push(sarg);
+			} else {
+				fn_args.push(arg);
+			}
+		}
+
+		let special_args: Vec<_> = special_args;
+		let fn_args: Vec<_> = fn_args;
```
tomusdrw, line 237:
```suggestion
		let (special_args, fn_args) = {	
				let mut special_args = vec![];
				let mut fn_args = vec![];

				for (i, arg) in args {
					if let Some(sarg) = Self::special_arg(i, arg.clone()) {
						special_args.push(sarg);
					} else {
						fn_args.push(arg);
					}
				}
				
				(special_args, fn_args)
		};
```
Instead of re-assigning we can scope the code in a block, imho it's a bit cleaner and indicates the purpose better.

derive/src/to_delegate.rs
```diff
@@ -206,21 +206,39 @@ impl RpcMethod {
...
+
+		let param_types: Vec<_> = fn_args.iter().cloned().map(|arg| *arg.ty).collect();
+		let arg_names: Vec<_> = fn_args.iter().cloned().map(|arg| *arg.pat).collect();
```
tomusdrw, line 240:
```suggestion
		let param_types = fn_args.iter().map(|arg| *arg.ty).collect::<Vec<_>>();
		let arg_names = fn_args.iter().map(|arg| *arg.pat).collect::<Vec<_>>();
```
My preference would be to add a generic parameter to `collect`, but I'm not super strong on this.

Also `.cloned()` doesn't seem required if we dereference/copy the type, is it?

derive/src/to_delegate.rs
```diff
@@ -232,28 +250,38 @@ impl RpcMethod {
...
+				},
 			}
 		};
 
 		let method_ident = self.ident();
 		let result = &self.trait_item.sig.output;
-		let extra_closure_args: &Vec<_> = &special_args.iter().cloned().map(|arg| arg.0).collect();
-		let extra_method_types: &Vec<_> = &special_args.iter().cloned().map(|arg| arg.1).collect();
+		let extra_closure_args: Vec<_> = special_args.iter().cloned().map(|arg| *arg.pat).collect();
+		let extra_method_types: Vec<_> = special_args.iter().cloned().map(|arg| *arg.ty).collect();
```
tomusdrw, line 284: same comments about `cloned()`, maybe disregard the previous suggestion to add generic parameter to `collect` to keep consistency.

derive/src/to_delegate.rs
```diff
@@ -301,34 +329,35 @@ impl RpcMethod {
...
-			}
-		});
-		let subscriber_arg = param_types.get(1).and_then(|ty| {
-			if let syn::Type::Path(path) = ty {
-				if path.path.segments.iter().any(|s| s.ident == SUBSCRIBER_TYPE_IDENT) {
-					Some(ty.clone())
-				} else {
-					None
+	fn special_arg(index: usize, arg: syn::PatType) -> Option<syn::PatType> {
+		match index {
```
tomusdrw, line 333: :+1:  much cleaner, thanks!

tomusdrw (approved):

derive/src/to_delegate.rs
luc65r: `syn::Type` and `syn::Pat` don't implement `Copy`, so there must be a `.clone()` somewhere.
It should be more efficient to clone only `syn::Type` or `syn::Pat` instead of the whole `syn::FnArg`.

PR #637, opened 2021-09-08
Enable tcp feature of hyper by default.

### LINKED ISSUE #636: Fail to enable `http` feature of `jsonrpc-core-client`
I enabled `http` feature of `jsonrpc-core-client` and I got this issue:

```
error[E0599]: no function or associated item named `new` found for struct `Client` in the current scope
  --> /home/yvan/.cargo/registry/src/github.com-1ecc6299db9ec823/jsonrpc-client-transports-18.0.0/src/transports/http.rs:32:23
   |
32 |     let client = Client::new();
   |                          ^^^ function or associated item not found in `Client<_, _>`
```

This occurs in an empty `fn main() {}` program, just by adding `jsonrpc-core-client = { version = "18.0.0", features = ["http"] }` line in `Cargo.toml` (you can found the full demo program here: https://github.com/yvan-sraka/jsonrpc-client-http-demo).

Due to the lack of documentation, I couldn’t determine if this behavior is normal (and I just didn’t find out yet how to use the feature) or if it shouldn’t fail this way?

I was using Rust nightly and I test that the issue didn’t occurs with previous version `17.0.0` of `jsonrpc-core-client` crate …

tomusdrw:
Resolves #636

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,7 +9,7 @@ trim_trailing_whitespace=true
 max_line_length=120
 insert_final_newline=true
 
-[{.travis.yml,appveyor.yml}]
+[{.travis.yml,appveyor.yml,.gitlab-ci.yml}]
 indent_style=space
 indent_size=2
 tab_width=8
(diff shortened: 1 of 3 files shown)


niklasad1 (approved):

niklasad1:
right, the `HttpConnector` is hidden behind the `tcp` feature TIL

dvdplm (approved):

Issue #212, opened 2017-10-29
Request<T> and Response<T> for faster serialization


faern:
The current `Request` and `Response` objects ultimately hold on to `serde_json::Value` for their actual values. This forces a user of those objects to first serialize/deserialize their types `T` to/from a `serde_json::Value` between the wire representation and the final `T` representation. I have not done any performance benchmarks, but I guess serializing twice does cost quite a lot of extra resources.

If these types had a generic it would be possible to have them deserialize directly from a wire format to their target type `T`. There would of course be some obstacles to overcome to make this work, but I think it could work. What do you think?

tomusdrw:
Yup, make perect sense. Probably can be even done with backward compatibility by doing `Request<T: DeserializeOwned = serde_json::Value>`

faern:
Possible problem is things like `Params::Array(Vec<Value>)` where each item in the vector could have a different type. At the moment the way I solve it in `jsonrpc-client-core` is that I let the `T` represent the tuple of all types. so If I have a `String` and `u64` as arguments to a method then `T` is `(String, u64)`.

It basically works because tuples are serialized with `[...]` so it becomes a list in json. Not completely sure if this can be generalized to the `Request` object or not. So we would rather need something like `Request<T: DeserializeOwned = jsonrpc_client_core::types::Params>`

PR #639, opened 2021-09-27
Performance: Avoid converting responses to serde_json::Value

ckamm:
Previously, all responses were converted to serde_json::Value before
being serialized to string. Since jsonrpc does not inspect the result
object in any way, this step could be skipped.

Now, result objects are serialized to string much earlier and the Value
step is skipped.

This patch has large performance benefits for huge responses: In tests
with 4.5 GB responses, the jsonrpc serialization overhead after
returning from an rpc function was reduced by around 35%. Which means
several seconds of speed-up in response times.

To accomplish this while mostly retaining API compatibility, the traits
RpcMethod, RpcMethodSync, RpcMethodSimple are now generic in their
return type and are wrapped when added to an io handler.

There's now a distinction between the parsed representation of jsonrpc
responses (Output/Response) and result of rpc functions calls
(WrapOutput/WrapResponse) to allow the latter to carry the
pre-serialized strings.

Review notes:
- I'm not happy with the WrapOutput / WrapResponse names, glad for
  suggestions.
- Similarly, rpc_wrap must be renamed and moved.
- The http handler health request is now awkward, and must extract the
  result/error from the already-serialized response. Probably worth it.
- The largest API breakage I could see is in the middleware, where the
  futures now return WrapOutput/WrapResult instead of Output/Result.
- One could make WrapOutput just be String, but having a separate type
  is likely easier to read.

See #212


core/src/io.rs
```diff
@@ -475,9 +480,33 @@ fn read_request(request_str: &str) -> Result<Request, Error> {
...
+			// Not elegant, but guaranteed to only copy responses once
+			let mut response = String::with_capacity(length);
+			response.push('[');
+			for output in outputs {
+				if response.len() > 1 {
+					response.push(',');
+				}
+				response.push_str(&output.response);
+			}
+			response.push(']');
```
maciejhirsz, line 505: Since you guarantee that there is always at least one element in the array, you can skip branching if you always push a comma at the end, and replace the last one with a bracket:

```suggestion
			for output in outputs {
				response.push_str(&output.response);
				response.push(',');
			}
			response.pop();
			response.push(']');
```

maciejhirsz:
You could avoid all the generics with `Box<RawValue>` using the [`to_raw_value`](https://docs.serde.rs/serde_json/value/fn.to_raw_value.html) function in place of the `Value`. `Box<RawValue>` is just a `Box<str>` that's guaranteed to be valid JSON, which `serde_json` copies as-is to whatever `Serialize`-able type you put it in. Compared to passing the result of the output with generics there is going to be extra allocation and memcpy (to put the JSON string of the result into the JSON string of the whole response), but that overhead might still be insignificant in the grand scheme of things.

ckamm:
> You could avoid all the generics with `Box<RawValue>` using the [`to_raw_value`](https://docs.serde.rs/serde_json/value/fn.to_raw_value.html) function in place of the `Value`. `Box<RawValue>` is just a `Box<str>` that's guaranteed to be valid JSON, which `serde_json` copies as-is to whatever `Serialize`-able type you put it in. Compared to passing the result of the output with generics there is going to be extra allocation and memcpy (to put the JSON string of the result into the JSON string of the whole response), but that overhead might still be insignificant in the grand scheme of things.

I had a similar idea, but my benchmarks showed that using `RawValue` instead of `Value` as an intermediate didn't actually save much time: 
```
to_value+to_string: total 956ms, to_value 344ms, to_string 612ms
to_rawvalue+to_string: total 943ms, to_rawvalue 594ms, to_string 349ms
to_string: total 593ms
clone: total 338ms
```
This was for a 1GB result. `clone` is just calling `String::clone` on a string of that length - suggesting that this may be memory bandwidth constrained.

maciejhirsz:
Yeah, that would be 3GB/s for a memcpy, which sounds about right for a payload of this size on a single thread. I would have expected `to_value` to be more expensive (and the whole optimization to be way more than 35% cut) as it converts all structs to HashMaps, but without seeing the data you are benchmarking on it's hard to tell what's happening.

ckamm:
Yeah, it's a very simple benchmark - can share it if you're interested. In the wild this has improved time-to-first-byte times from around 26s to 19s for 4.5 GB responses. The time not spent in the actual rpc handler function has gone down from 14s to 8s. (of course having responses of that size it a problem in itself)

ckamm:
I've done the renames.

Do you have an idea for what to do about `rpc_wrap()`. Should I implement `Into/From` for the `RpcMethod -> RpcMethodWithSerializedOutput` conversion, or keep it explicit?

core/src/calls.rs
```diff
@@ -61,11 +63,24 @@ pub trait RpcNotification<T: Metadata>: Send + Sync + 'static {
 	fn execute(&self, params: Params, meta: T);
 }
 
+pub trait RpcMethodWithSerializedOutput<T: Metadata>: Send + Sync + 'static {
+	fn call(&self, params: Params, meta: T, jsonrpc: Option<Version>, id: Id) -> BoxFuture<Option<SerializedOutput>>;
+}
+
+pub fn rpc_wrap<T: Metadata, R: Serialize + Send + 'static, F: RpcMethod<T, R>>(
```
tomusdrw, line 70:
```suggestion
pub(crate) fn rpc_wrap<T: Metadata, R: Serialize + Send + 'static, F: RpcMethod<T, R>>(
```

I think this can stay crate-local.

core/src/calls.rs
```diff
@@ -61,11 +63,24 @@ pub trait RpcNotification<T: Metadata>: Send + Sync + 'static {
 	fn execute(&self, params: Params, meta: T);
 }
 
+pub trait RpcMethodWithSerializedOutput<T: Metadata>: Send + Sync + 'static {
+	fn call(&self, params: Params, meta: T, jsonrpc: Option<Version>, id: Id) -> BoxFuture<Option<SerializedOutput>>;
+}
```
tomusdrw, line 68: Could you add some docs here?

tomusdrw (approved):
lgtm!

core/src/calls.rs
```diff
@@ -61,11 +63,26 @@ pub trait RpcNotification<T: Metadata>: Send + Sync + 'static {
 	fn execute(&self, params: Params, meta: T);
 }
 
+/// Asynchronous Method with Metadata, returning a serialized JSONRPC response
```
tomusdrw, line 66:
```suggestion
/// Asynchronous Method with Metadata, returning a serialized JSON-RPC response.
```

core/src/calls.rs
```diff
@@ -61,11 +63,26 @@ pub trait RpcNotification<T: Metadata>: Send + Sync + 'static {
...
+pub trait RpcMethodWithSerializedOutput<T: Metadata>: Send + Sync + 'static {
+	fn call(&self, params: Params, meta: T, jsonrpc: Option<Version>, id: Id) -> BoxFuture<Option<SerializedOutput>>;
```
tomusdrw, line 68:
```suggestion
    /// Executes the RPC method and returns [`SerializedOutput`].
	fn call(&self, params: Params, meta: T, jsonrpc: Option<Version>, id: Id) -> BoxFuture<Option<SerializedOutput>>;
```

tomusdrw (approved):

PR #640, opened 2021-10-02
Omit params field from requests instead of sending null value

selu:
Based on the [specification](https://www.jsonrpc.org/specification#parameter_structures), the request's params must be an array or an object if it exists, so sending a null value is not valid. Actually, it causes errors for KODI JSONRPC server, that's why I tried to find a way to fix it.

The solution is very simple, just I haven't updated the comments for the params field.

### DIFF
--- a/core/src/types/request.rs
+++ b/core/src/types/request.rs
@@ -12,7 +12,7 @@ pub struct MethodCall {
 	pub method: String,
 	/// A Structured value that holds the parameter values to be used
 	/// during the invocation of the method. This member MAY be omitted.
-	#[serde(default = "default_params")]
+	#[serde(default = "default_params", skip_serializing_if = "Params::is_none")]
 	pub params: Params,
 	/// An identifier established by the Client that MUST contain a String,
 	/// Number, or NULL value if included. If it is not included it is assumed
--- a/core/src/types/params.rs
+++ b/core/src/types/params.rs
@@ -39,7 +39,6 @@ impl Params {
 
 	/// Check for None
 	pub fn is_none(&self) -> bool {
-		log::debug!("is_none called on {:?}", self);
 		matches!(*self, Params::None)
 	}
 }
--- a/core/src/types/params.rs
+++ b/core/src/types/params.rs
@@ -36,6 +36,12 @@ impl Params {
 			p => Err(Error::invalid_params_with_details("No parameters were expected", p)),
 		}
 	}
+
+	/// Check for None
+	pub fn is_none(&self) -> bool {
+		log::debug!("is_none called on {:?}", self);
+		matches!(*self, Params::None)
+	}
 }
 
 impl From<Params> for Value {


core/src/types/request.rs
```diff
@@ -105,6 +105,21 @@ mod tests {
 		);
 	}
 
+	#[test]
+	fn call_serizalize_without_params() {
```
tomusdrw, line 109:
```suggestion
	fn call_serialize_without_params() {
```

tomusdrw (commented):
@selu thanks a lot! This looks indeed sensible. Do you mind adding an additional test case to make sure that the server side of the library is correctly converting missing `params` into `None`?

I think adding a reverse test case (i.e. deserialize without params) would suffice, unless I didn't see we already have a testcase like that.

core/src/types/request.rs
selu: Yes, sure, I'll add that test case. Probably, I can do it tomorrow.

niklasad1 (commented):
Thanks, I agree with your interpretation of the spec.

I suggest an additional test for the `jsonrpc derive` (if possible):

1. a method without params: `fn a() -> BoxedFuture`
2. a method with optional param: `fn b(p: Option<usize>)`

selu:
@tomusdrw , I found the test case for deserialize without params, it is the last item of `fn call_deserialize()` between lines 263 and 273 in file core/src/types/request.rs

selu:
> Thanks, I agree with your interpretation of the spec.
> 
> I suggest additional tests for `jsonrpc derive` (if possible):
> 
>     1. a method without params: `fn a() -> BoxedFuture`

There is already a test case for empty params in derive/tests/macros.rs
```rust
#[rpc]
pub trait Rpc {
	/// Returns a protocol version.
	#[rpc(name = "protocolVersion")]
	fn protocol_version(&self) -> Result<String>;
…
fn should_accept_empty_array_as_no_params() {
	let mut io = IoHandler::new();
	let rpc = RpcImpl::default();
	io.extend_with(rpc.to_delegate());

	// when
	let req1 = r#"{"jsonrpc":"2.0","id":1,"method":"protocolVersion","params":[]}"#;
	let req2 = r#"{"jsonrpc":"2.0","id":1,"method":"protocolVersion","params":null}"#;
	let req3 = r#"{"jsonrpc":"2.0","id":1,"method":"protocolVersion"}"#;
```
However, this checks another cases too.
> 
>     2. a method with optional param: `fn b(p: Option<usize>)`

I could not find test case for optional param, probably it worth it to add one. Would it be ok in the same file in a similar way?

niklasad1:
> I could not find test case for optional param, probably it worth it to add one. Would it be ok in the same file in a similar way?

Yes, just make a new unit test for it in that file should be fine

selu:
@niklasad1 , test is added for the optional parameter, please review.

derive/tests/macros.rs
```diff
@@ -222,6 +231,47 @@ fn should_accept_any_raw_params() {
...
 }
 
+#[test]
+fn should_accept_optional_param() {
+	let mut io = IoHandler::new();
+	let rpc = RpcImpl::default();
+	io.extend_with(rpc.to_delegate());
+
+	// when
+	let req1 = r#"{"jsonrpc":"2.0","id":1,"method":"sqr","params":[2]}"#;
```
niklasad1, line 241: I would like to add `[]` and `null` here otherwise looks great

derive/tests/macros.rs
```diff
@@ -222,6 +231,49 @@ fn should_accept_any_raw_params() {
...
+	let req2 = r#"{"jsonrpc":"2.0","id":1,"method":"sqr","params":[]}"#;
+	let req3 = r#"{"jsonrpc":"2.0","id":1,"method":"sqr","params":null}"#;
```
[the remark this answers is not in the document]
tomusdrw: > hrm, really? In the client code Params::None would rejected then?

I was checking the code and it might not be what we do now. AFAICT in case of optional parameters we parse `None` correctly as well. Anyways, I think this discussion is a bit tangential to that particular PR, so let's open an issue and discuss there.

The code in PR is an improvement for the client side for sure, we might want to follow up on how we want the server to behave separately.

Issue #643, opened 2021-10-07
[ws server]: configuration `ServerBuilder::max_payload` causes unexpected allocations


niklasad1:
When I ran an example `WS server`:

```rust
#[tokio::main]
async fn main() {
    use jsonrpc_ws_server::jsonrpc_core::*;
    use jsonrpc_ws_server::*;

    let _ = env_logger::try_init();

    let mut io = IoHandler::new();
    io.add_method("say_hello", |_| async {
        Ok(Value::String("lo".to_string()))
    });

    let server = ServerBuilder::new(io)
        .max_payload(usize::MAX)
        .start(&"127.0.0.1:3030".parse().unwrap())
        .expect("Server must start with no issues");

    server.wait().unwrap()
}
```

Then when I trying to connect to the server, the server crashed with the error `memory allocation of 22517998136852480 bytes failed`... I expected this to occur only when sending huge amount of data to the server not just by connecting....

//cc @tomusdrw @koute

I think this is related to the fix with memory leak in `parity-ws 0.11` but a pure guess from my side.

PR #644, opened 2021-10-08
Limit the maximum WebSocket fragments capacity

koute:
Fixes https://github.com/paritytech/jsonrpc/issues/643

If `max_payload_bytes` is very big this also sets the `fragments_capacity` to a very big number, and the `parity-ws` crates uses `fragments_capacity` to preallocate a `VecDeque` on each connection, which then fails.

So here we add a reasonable-ish limit so that the `fragments_capacity` isn't *too big*. Considering the `config.fragment_size`'s default value is 64k we'd need a `max_payload_bytes` of 256MB to reach it, which should be plenty. (Most likely no one should set it this high anyway due to the possibility of leaving the server open to a denial of service attack.)

### DIFF
--- a/ws/src/server.rs
+++ b/ws/src/server.rs
@@ -70,6 +70,7 @@ impl Server {
 			config.max_connections = max_connections;
 			// don't accept super large requests
 			config.max_fragment_size = max_payload_bytes;
+			config.max_total_fragments_size = max_payload_bytes;
 			config.in_buffer_capacity_hard_limit = max_in_buffer_capacity;
 			config.out_buffer_capacity_hard_limit = max_out_buffer_capacity;
 			// don't grow non-final fragments (to prevent DOS)
--- a/ws/src/server.rs
+++ b/ws/src/server.rs
@@ -75,6 +75,10 @@ impl Server {
 			// don't grow non-final fragments (to prevent DOS)
 			config.fragments_grow = false;
 			config.fragments_capacity = cmp::max(1, max_payload_bytes / config.fragment_size);
+			if config.fragments_capacity > 4096 {
+				config.fragments_capacity = 4096;
+				config.fragments_grow = true;
+			}
 			// accept only handshakes beginning with GET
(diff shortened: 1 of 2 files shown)


ordian (approved):
Sounds reasonable.

ws/src/server.rs
```diff
@@ -74,7 +74,7 @@ impl Server {
 			config.out_buffer_capacity_hard_limit = max_out_buffer_capacity;
 			// don't grow non-final fragments (to prevent DOS)
 			config.fragments_grow = false;
-			config.fragments_capacity = cmp::max(1, max_payload_bytes / config.fragment_size);
+			config.fragments_capacity = cmp::min(cmp::max(1, max_payload_bytes / config.fragment_size), 4096);
```
tomusdrw, line 77: Isn't 4096 too little? Since fragments cannot grow, with 4096 fragments and `fragment_size` set to default we will be able to handle fragmented payload size up to `4096 * 65535 ~= 255MB`. 
Maybe better change `max_payload_bytes` to be an optional parameter (`None` meaning no limit and growing fragments) and have it fail early with a helpful error message in case `max_payload_bytes` is unreasonably big (bigger than the amount of RAM or sth)?

tomusdrw: I'm not arguing about sensibility of the limit. I agree that for our use cases 256MB of request size is probably more than enough.
I am against making Polkadot/Substrate specific assumptions in the general `jsonrpc` library we are building here. Silently capping the max amount of bytes in a request, despite having a separate configuration option (`max_payload_bytes`) is IMHO incorrect.
I'd suggest to simply error out if `max_payload_bytes` is configured for something above `255MB` and keeping the behavior introduced in this PR or (IMHO more in the vein of "general library") making it more flexbile by enabling growing fragments.

koute: You do make a good point.

Okay, so I've made it so that if the fragments' capacity is over the limit it just lets them grow.

Issue #642, opened 2021-10-07
Deserialization of `params: null`


tomusdrw:
Currently `params: null` in JSON will be deserialized into `Params::None`. I'm not entirely sure if this behavior is expected by the spec. If we want to be very strict, we could instead enforce that the `params` field is not present in case there is no parameters to pass.

Issue #595, opened 2020-10-25
Option to drop `deny_unknown_fields` annotations


jez:
This feature request is a follow up to this PR:

https://github.com/paritytech/jsonrpc/pull/448

Specifically, there was a comment:

> I like strict more... jsonrpc provides a way to send arbitrary data in a params field, so I don't see a reason why someone would need to use other fields

While there is the `params` field for requests, for responses there is just `result` or `error`.

The use case that I have for extra top-level fields is that in some languages it's really annoying to have to deserialize untagged unions. The easy fix is to add an extra tag, that can then be ignored by tools that don't need to look for it. If I ever want to look at a JSON RPC response that I know my code serialized, I can look for that extra field, treat it as a tag, and then deserialize according to the value of that tag.

I would like to urge you all to add this back as an option. We would love to make use of it in LanguageClient-neovim. Language Server Protocol clients want to be as permissive as possible to accommodate as many language servers as possible (consider: we can't fix all the language servers, so we try to accomodate them).

A flag to optionally skip the `deny_unknown_fields` would be appreciated.

tomusdrw:
Related: https://github.com/paritytech/jsonrpc/pull/649

The `non-strict` mode could disregard members casing (but not method casing).

PR #649, opened 2021-11-06
Allow capital case on JSONRPC message

MrCasCode:
Some providers send a non-standard typing, so add an alias to allow decoding capital version

### DIFF
--- a/core/src/types/request.rs
+++ b/core/src/types/request.rs
@@ -7,6 +7,7 @@ use super::{Id, Params, Version};
 #[serde(deny_unknown_fields)]
 pub struct MethodCall {
 	/// A String specifying the version of the JSON-RPC protocol.
+	#[serde(alias = "JSONRPC")]
 	pub jsonrpc: Option<Version>,
 	/// A String containing the name of the method to be invoked.
 	pub method: String,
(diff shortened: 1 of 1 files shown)


tomusdrw:
Hey @MrCasCode, thank you for the contribution. We are trying to implement the JSON-RPC specification strictly (see also https://github.com/paritytech/jsonrpc/issues/595, https://github.com/paritytech/jsonrpc/pull/448, https://github.com/paritytech/jsonrpc/issues/642) and [the spec](https://www.jsonrpc.org/specification) states:

> All member names exchanged between the Client and the Server that are considered for matching of any kind should be considered to be case-sensitive. The terms function, method, and procedure can be assumed to be interchangeable.

Hence I'm going to close the PR without merging - adding support for arbitrary implementations that don't match the spec is not something we want to support long term - feel free to reach out to the providers you've mentioned and have them fix the member case. 

Alternatively please consider implementing a more generic solution of `non-strict` mode, given in such mode we could support `JSONRPC` member I don't see a reason why `JSONrpc` or `JsOnRpC` should be rejected - i.e. we should ignore casing in members and also allow arbitrary extra members as described in #595.

I hope you'll understand the decision, but feel free to re-open or discuss further.

PR #650, opened 2021-11-14
move to rust 2021, macro tweak to compile

NukeManDan:
Confirmed to build **and pass tests** locally on Ubuntu 20.04.3 LTS:
```
rustc 1.56.1 (59eed8a2a 2021-11-01)
rustc 1.58.0-nightly (e90c5fbbc 2021-11-12)
```

### DIFF
--- a/core-client/transports/Cargo.toml
+++ b/core-client/transports/Cargo.toml
@@ -1,14 +1,15 @@
 [package]
+name = "jsonrpc-client-transports"
+version = "18.0.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 description = "Transport agnostic JSON-RPC 2.0 client implementation."
 documentation = "https://docs.rs/jsonrpc-client-transports/"
-edition = "2018"
 homepage = "https://github.com/paritytech/jsonrpc"
 keywords = ["jsonrpc", "json-rpc", "json", "rpc", "serde"]
 license = "MIT"
-name = "jsonrpc-client-transports"
 repository = "https://github.com/paritytech/jsonrpc"
-version = "18.0.0"
+edition = "2021"
+rust-version = "1.56.1"
 
 categories = [
 	"asynchronous",
--- a/derive/src/to_client.rs
+++ b/derive/src/to_client.rs
@@ -58,12 +58,12 @@ pub fn generate_client_module(
 
 			/// The Client.
 			#[derive(Clone)]
-			pub struct Client#generics {
+			pub struct Client #generics {
 				inner: TypedClient,
 				#(#markers_decl),*
 			}
 
-			impl#generics Client#generics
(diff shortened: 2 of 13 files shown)


derive/src/to_client.rs
```diff
@@ -78,7 +78,7 @@ pub fn generate_client_module(
 				#(#client_methods)*
 			}
 
-			impl#generics From<RpcChannel> for Client#generics
+			impl #generics From<RpcChannel> for Client #generics
```
KiChjang, line 81: What would happen if you did not make these whitespace changes?

bkchr: That is done by cargo fix.

See: https://doc.rust-lang.org/edition-guide/rust-2021/reserving-syntax.html

NukeManDan:
`rust-version = "1.56.1"` introduced

For reference, I tried to get this to `cargo build` with 1.56.0 and it failed with
```
error: package `...` cannot be built because it requires rustc 1.56.1 or newer, while the currently active rustc version is 1.56.0
```
With `--ignore-rust-version` it _does_ build with 1.56.0, as expected.

KiChjang (approved):

core-client/transports/Cargo.toml
```diff
@@ -1,14 +1,15 @@
 [package]
+name = "jsonrpc-client-transports"
+version = "18.0.0"
```
tomusdrw, line 3: We should probably bump the major versions everywhere due to different compiler requirements.

tomusdrw (approved):

core-client/transports/Cargo.toml
NukeManDan: do you  mean minor or patch? 18.1.0 makes sense to me, as its compatible with 18.0.0 
maybe even just 18.0.1 as this doesn't effect the behavior of the program

tomusdrw: I actually really meant `major`. The thing is that we now suddenly require everyone to use a specific Rust compiler version which IMHO is a breaking change. But I'm not a semver expert, maybe @dvdplm has some suggestions.

dvdplm: MSRV changes are generally considered breaking. v19! :)

NukeManDan: Oh well, I can do that here, sure! But I know that many do not do this (I put in a lot of PRs across parity for this, and it so far has not been)
Happy to update across the board and reference this thread to justify. 

Do note that it's not universally agreed that MSVR bump = breaking change. 
- https://github.com/libp2p/rust-libp2p/pull/2339#issuecomment-969089617
- https://github.com/libp2p/rust-libp2p/pull/2339#issuecomment-971316894

PR #653, opened 2021-12-09
Do not pull in url-1.0 unless websocket is enabled

notriddle:
This takes advantage of rust-websocket re-exporting the `url` crate, making this easy.

### DIFF
--- a/core-client/transports/Cargo.toml
+++ b/core-client/transports/Cargo.toml
@@ -48,7 +48,7 @@ hyper-tls = { version = "0.5", optional = true }
 jsonrpc-server-utils = { version = "18.0.0", path = "../../server-utils", optional = true }
 parity-tokio-ipc = { version = "0.9", optional = true }
 tokio = { version = "1", optional = true }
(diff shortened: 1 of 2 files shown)


tomusdrw (approved):

tomusdrw (commented):
Might actually be good to bump the `websocket` dependency as well, the one we use is 2 years old :)

notriddle:
If you really want that, sure, but the current version of rust-websocket still hasn't upgrade to url-2.0, so it wouldn't really matter to my purposes (I just wanted to clean up the transitive dependencies of a project that doesn't even use the websocket transport :smile:)

tomusdrw:
Yeah, fair enough, we don't use strict dependency here, so I guess it serves as a minimal version requirement here which is probably fine.

notriddle:
@tomusdrw Okay, I've bumped websocket to 0.26. The old version, 0.24, was not a compatible version number.

niklasad1 (approved):

notriddle:
Is anything else blocking this?

PR #656, opened 2022-01-10
Remove returns strings

Cupnfish:
Removes parameters that are not necessary.
Disadvantages:

- The `type_name` method does not currently have stable `const`.
- The value returned by `type_name` is not necessarily the shortest string, for example `type_name::<Option<String>>()` may return `Option<String>` or `std::option::Option<std::string::String>`.
- Also the returned value will vary with different versions of the compiler.

Advantages:

- Reduces unnecessary parameter input, which when calling a method will make the caller believe that the `return` value specified within the parameter has some relationship to the output `return` value, when in fact the `return` value entered within the parameter is only useful as a diagnostic for error output. The documentation of `type_name` within std, on the other hand, is meant to be used for code diagnostic purposes.
- The interface to call would be more clean.

### DIFF
--- a/core-client/transports/src/lib.rs
+++ b/core-client/transports/src/lib.rs
@@ -289,7 +289,8 @@ impl TypedClient {
 
 			log::debug!("response: {:?}", value);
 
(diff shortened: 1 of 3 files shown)


niklasad1 (approved):
nice one, LGTM

PR #658, opened 2022-01-18
Fix core examples so they don't panic

marsella:
The `basic` and `async` examples did not correctly encode the expected response. Now they do!


niklasad1 (approved):

marsella:
Can I get a second review? Not sure if @tomusdrw is the best person - please tag someone else if appropriate.

tomusdrw (approved):

tomusdrw:
Thanks!

PR #670, opened 2022-04-11
Make windows http server multithreaded

i1i1:
This pr cleans up some old hacky code which implied older version of mio and tokio. This would make HTTP server on windows be multithreaded.


jsdw (approved):
LGTM, and thank you!

niklasad1 (approved):

i1i1:
Btw, could you also update library on crates.io? It would be great to pull this patch from there :smile:

PR #677, opened 2022-07-26
add nonstrict feature to jsonrpc-core

### LINKED ISSUE #595: Option to drop `deny_unknown_fields` annotations

clouds56:
With nonstrict, we do not deny any unknown field, instead we collect them into `other` field.
This maybe fixes #595

### DIFF
--- a/core-client/transports/src/transports/http.rs
+++ b/core-client/transports/src/transports/http.rs
@@ -249,10 +249,7 @@ mod tests {
 		if let Err(RpcError::JsonRpcError(err)) = res {
 			assert_eq!(
 				err,
-				Error::new_with_message(
-					ErrorCode::ServerError(-34),
-					"Server error"
-				)
+				Error::new_with_message(ErrorCode::ServerError(-34), "Server error")
 			)
 		} else {
 			panic!("Expected JsonRpcError. Received {:?}", res)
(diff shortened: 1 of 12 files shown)


niklasad1:
Hey, it's just `cargo fmt` that fails so run it on your code and it should make the CI happy

perekopskiy:
Hi, if this PR is going to lend eventually, I'd like to ask you to remove `#[serde(deny_unknown_fields)]` with non-strict mode not only for the responses but for the requests too. The reason is some clients include non-standard fields and some servers need to maintain it.

niklasad1:
hey again,

we are not maintaining this library actively anymore, https://github.com/paritytech/jsonrpsee also supports this

